PR #6, opened 2021-05-31
Typescript and fixes to NextJS Documentation

gusfune:
This contribution adds 2 main things to the NextJS documentation.

1.  Replacing let with const on NextJS docs fdae022
ESLint warns for using `let` in a variable that is not being changed later. In this case `const` is the more appropriate definition.

2. Added Typescript for first steps in NextJS a27ca87
Current NextJS documentation is only for JS, added a few details to make it compatible with Typescript.


rishabhpoddar (approved):

rishabhpoddar:
Thanks for this pr @gusfune. I have also added you on the readme for supertokens-core :)

PR #43, opened 2021-07-01
Add SessionAuth and other session wrappers docs

dulowski-marek:
### Related:
- https://github.com/supertokens/supertokens-core/issues/269
- https://github.com/supertokens/supertokens-auth-react/pull/274
- https://github.com/supertokens/supertokens-website/pull/64

## TODO:
- [x] Organise thoughts on docs
- [x] Make changes
- [x] New version of auth-react docs (need to update the installation command as well(


auth-react/docs/emailpassword/email-password-auth.md
```diff
@@ -32,4 +32,5 @@ __HIGHLIGHT__        return (
        userId: string,
        jwtPayload: any /*JSON object set on the backend*/
    }
-   ```
\ No newline at end of file
+   ```
+- `EmailPasswordAuth` will update the session, because it uses `SessionAuth` underneath. Read more about session updates in [SessionAuth guide](/docs/auth-react/docs/session/session-auth).
```
rishabhpoddar, line 36: When linking to pages within the same docusaurus project, use relative path

rishabhpoddar, line 36: You also want to mention that this takes an `onSessionExpired` prop?

emailpassword/docs/common-customizations/sessions/user-information-front-end.md
```diff
@@ -49,6 +49,7 @@ render() {
 }
 ```
 
+You can read more about `EmailPasswordAuth` in [its API guide](/docs/auth-react/docs/emailpassword/email-password-auth)
```
rishabhpoddar, line 52: I suggest that you understand what the current recipe docs say about these auth wrappers, and where all it's mentioned.

Then you should think about how best to explain everything about these auth wrappers. You want to talk about:
- Protecting a website route
- Checking if a user is logged in
- Fetching session info on the frontend (this page)
- Handling session expiry
- Anything else?

Once you have a clear picture of what needs to be communicated, then you can see the current pages that exist. If you can add the content to the current pages itself, then cool, else add new pages.

If you'd like to remove current pages and rewrite them entirely, that's an option too.

rishabhpoddar (changes requested):

emailpassword/docs/common-customizations/sessions/user-information-front-end.md
dulowski-marek: Exactly those usecases are covered in `SessionAuth` docs. I think the usecases you mentioned should be described in `auth-react` docs, since like right now we added a feature, which added one usecase (session expiry). 

I can revisit this page to contain the usecases, but it's going to differ between SDK versions.

One solution I can think about is just mention few of them (protecting website route/component, providing session data to child components) and link to the API.

auth-react/docs/emailpassword/email-password-auth.md
dulowski-marek: I can duplicate them from `SessionAuth`, yes. It definitely makes sense.

emailpassword/docs/common-customizations/sessions/user-information-front-end.md
rishabhpoddar: > One solution I can think about is just mention few of them (protecting website route/component, providing session data to child components) and link to the API.

This works. The way we have structured our docs right now is that all "guides" / "how tos" are in the recipe docs, and the SDK docs are only to see what the available functions are.

auth-react/docs/emailpassword/email-password-auth.md
dulowski-marek: I wanted to avoid doing `../` here, but this is only preference. Both styles will cause it to break link if something moves, so if you want me to change to `../`, I will.

rishabhpoddar: The reason I insist is that cause in v2 docusaurus (which we are working on moving towards), using relative path doesn't cause a reload of the whole docs..

dulowski-marek: And that's a perfect reason, thanks for explaining! I'll change that.

emailpassword/docs/common-customizations/sessions/securing-component.md
```diff
@@ -0,0 +1,25 @@
+---
+id: securing-component
+title: Securing component
+hide_title: true
+---
+
+# Securing a component
```
rishabhpoddar, line 7: This title is very confusing.. something like "Securing a website route" is more clear

emailpassword/docs/common-customizations/sessions/user-information-front-end.md
```diff
@@ -49,6 +49,17 @@ render() {
 }
 ```
 
+### Automatic context updates
+`SessionContext` will be updated for `EmailPasswordAuth` children whenever any of the following events happens:
```
rishabhpoddar, line 53: This is too much information provided in this recipe docs.. it's good for auth-react SDK docs.. but in recipe, we don't need this.

session/docs/advanced-customizations/frontend-hooks/handle-event.md
```diff
@@ -20,6 +20,11 @@ Session.init({
             // NOTE: This is an undeterministic event
         } else if (context.action === "UNAUTHORISED") {
             // called when the session has expired
+        } else if (context.action === "SESSION_CREATED") {
+            // Called when session is created.
+            //
+            // It means that if session previously existed
+            // and it only gets refreshed, this event will not be fired
```
rishabhpoddar, line 27: Don't need this extra two lines. It adds confusion I think.

rishabhpoddar:
# Overall feedback:
- I think the lesser non essential details you give in the recipe docs, the better it is
- You seem to be forgetting to add "a", "an", "the" in sentences which make them grammatically incorrect.
- You forgot to change the version of supertokens-website SDK docs (which I have done already)
- Pay extra attention to titles of pages: "Securing a component" is not a clear title. Also, it's placement in the huge list of session related "how to's" is not the best. It should be higher up in the list I feel. But I have left it there for now.

Apart from the above, it's all good.

rishabhpoddar (approved):

Issue #8, opened 2021-06-02
Migrate to docusaurus v2


rishabhpoddar:
## TODO:
- [x] svg supertokens logo icon: https://supertokens.slack.com/archives/DRZDA5CUC/p1626092146006400
- [x] navbar logo resizing
- [x] driftt chat
- [x] Overall theme styling and colours
- [x] Change contributing readme in docs
- [x] Create latest versions of all the docs
- [ ] Algolia search is isolated properly per recipe and per version (maybe we can't add tags..).
- [x] Recipe name in recipe docs
- [x] Linking to older versions of the docs
- [x] Multi docs instance
- [x] Sub tabs styling
- [x] Meta tags for SEO in <head>
- [x] multiple instances of docs
- [x] footer: https://github.com/supertokens/docs/blob/docusaurus-v2-feasibility/community/supertokens-docs-plugin/index.js
   - Use plugin to add footer?
- [x] build process.
- [x] clicking on header buttons should take you to the right place
- [x] making community docs
   - [x] checking if redirection happens properly for certain docs
   - [x] hiding items in the sidebar
   - [x] getting stared page embed
   - [x] compatibility table
- [x] copy docs
- [x] analytics: https://github.com/supertokens/docs/blob/docusaurus-v2-feasibility/community/src/theme/Root.js
- [x] Is `TO_CHANGE.md` still relevant?
- [x] Check edit links
- [x] Check SEO meta tags + make sure og:url and twitter:url change on navigation + check social preview
- [x] remove docs meta data from sheetsdb
- [x] remove old recipe docs
- [x] Do a global search for `__END_HIGHLIGHT__` or `__` and other old ways of highlighting code and make sure they are all gone
- [x] Do a global search on `<!--DOCUSAURUS_CODE_TABS-->` and remove them.

rishabhpoddar:
v2 supprts code highlighting out of the box, so we can also get rid of our code highlighting method.

rishabhpoddar:
In V1, we hack around with some JS to make a top level side bar item into a page. Is that possible in v2 without the same hack?

rishabhpoddar:
@mufassirkazi please create a supertokens logo thats on the top left of our docs in .svg format.

mufassirkazi:
https://supertokens.slack.com/archives/DRZDA5CUC/p1626092146006400

rishabhpoddar:
@mufassirkazi , I need equivalent colours for the below, but for our colour scheme:
```
:root {
[...]

PR #65, opened 2021-07-18
Docusaurus v2

rishabhpoddar:
Related to https://github.com/supertokens/docs/issues/8


mufassirkazi:
I'll suggest we do 3 - (the theme name is ultramin). Even though the 2nd option was what I preferred, but UX wise I think the 3rd option is better.

https://zpl.io/29dwDpd

So if we are going with the third option, can you please change the background colors of the tabs and sub tabs as reflected in the design above?

rishabhpoddar:
@mufassirkazi ultramin is white theme..

PR #106, opened 2021-09-01
Added documentation for adding styles that apply across recipes

porcellus:
## Summary of change

Added documentation for an extra property passed to `SuperTokens.init` that applies styling to all components.

## Related issues
- supertokens/supertokens-auth-react#312

### DIFF
--- a/v2/emailpassword/common-customizations/styling/changing-style.mdx
+++ b/v2/emailpassword/common-customizations/styling/changing-style.mdx
@@ -12,10 +12,9 @@ import TabItem from '@theme/TabItem';
 
 Updating the CSS allows you to change the UI of our components to meet your needs.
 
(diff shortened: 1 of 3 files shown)


v2/emailpassword/common-customizations/styling/changing-style.mdx
```diff
@@ -66,6 +66,29 @@ The above will result in:
 
 <img width="500px" src="/img/emailpassword/signup-centered-blue-button.png" />
 
+### Customizing common components
```
rishabhpoddar, line 69: - The title is not clear - perhaps call it "Global style changes"?
- The placement of this section is not good - It should probably be it's own "##" section at the top of this file (as the first section), and then all the other sections will be specific to each form.

(Same comment applies to other recipes as well)

rishabhpoddar (changes requested):

PR #107, opened 2021-09-05
docs: clarity on axios interceptors

cerino-ligutom:
## Summary of change

Added clarity on the usage of the axios interceptors in the description.

Current description gives the wrong impression that it should always be set on the axios import but this doesn't work if we're not directly using the imported static axios object such as when [creating an axios instance](https://github.com/axios/axios#creating-an-instance) and using that instance throughout the codebase instead of importing axios everywhere.


rishabhpoddar:
Thanks for this PR @cerino-ligutom. I have also added your profile in the supertokens-core readme :)

Issue #102, opened 2021-08-30
Docs improvement 5


jscyo:
- [x] How to change sign up / in UI labels on the frontend
- [x] Add about changing OAuth scopes in thirdparty and thirdpartyemailpassword recipe
- [x] Add about getting the access token of the thirdparty provider in thirdparty and thirdpartyemailpassword recipes.
- [x] In sing up form > adding custom fields > the last section, give an example of how to fetch the field in the backend and don't just link to the post sign up callback.
- [x] In `CORS` section in backend quick setup, say that you need to add full website URL.

rishabhpoddar:
@jscyo can remove  “About this recipe” talks about a recipe - but what recipe is that? it’s not clear

jscyo:
- Create a separate page for  "Changing Field labels"
- Create a separate page for Changing OAuth scopes, give an example
- Fix title "Get the Third Party Providers Access Token" -> "Get the Third Party Provider's Access Token"
- Add a comment in thirdparty provider access token example saying that this is for google.
- Example in CORS should be https

rishabhpoddar:
- In the sign in label change thing, the screenshot says sign in, but the code is for the `signUpForm`. Also, you need to have a similar page for the sign in form sub section (for thirdpartyemailpassword and emailpassword)
- Change "When adding origins make sure to add the full website URL, for example, if the domain is example.com make sure the value set is https://example.com" to "When adding origins make sure to add the full website URL. For example, if the domain is `example.com` make sure the value set is `https://example.com`" (agains the `Access-Control-Allow-Origin` header)

PR #109, opened 2021-09-06
Doc improvement 5

jscyo:
## Summary of change
Docs for forontend UI label change, adding OAuth scopes, getting third-party accesstoken, handling formfields on the backend and CORS requires for website URL.

## Related issues
#102 

none


jscyo:
- Create a separate page for  "Changing Field labels"
- Create a separate page for Changing OAuth scopes, give an example
- Fix title "Get the Third Party Providers Access Token" -> "Get the Third Party Provider's Access Token"
- Add a comment in thirdparty provider access token example saying that this is for google.
- Example in CORS should be https

Issue #104, opened 2021-08-30
Add documentation for NestJS integration


porcellus:
Add step-by-step documentation to help devs use SuperTokens in Nest.
See NextJS docs for example.

PR #116, opened 2021-09-12
Added a guide on NestJS integration

porcellus:
## Summary of change
Added documentation about integrating SuperTokens with NestJS

## Related issues
- #104 

- [x] Add documentation into tabbed sections


rishabhpoddar:
- [ ] Try to add NextJS and NestJS logo on the left index to differentiate the two
- [ ] Add a NestJS sub tab in backend quick setup that takes the user to the nestjs section
   - [ ] Maybe link to the nestjs / nextjs sections in those seub tabs' default views.
- [ ] Remove the `\\`` from comments in certain places.

porcellus:
This is how the logos ended up: 
image
Also, right now they're linked from their respective sites, if you prefer I can download the logos and add them to the assets.

porcellus:
> Maybe link to the nestjs / nextjs sections in those seub tabs' default views.
It could be done, but the default tabs would need to be aware of the recipe (absolute link) or the path of the page (relative link).

rishabhpoddar:
> right now they're linked from their respective sites, if you prefer I can download the logos and add them to the assets.

Yes. Please download and put in our assets. Also, since you are doing this, could you also do the same for AWS lambda and netlify.


> It could be done, but the default tabs would need to be aware of the recipe (absolute link) or the path of the page (relative link).

Hmmm. Absolute links would break across recipes, and relative links would break across pages (since we are displaying the same component in several pages). So let's drop the idea of linking to it and keep it the way it is right now.

PR #126, opened 2021-09-26
Adding a guide to integrate SuperTokens with AWS AppSync

porcellus:
## Summary of change
Added a guide that shows how to integrate SuperTokens with AWS AppSync

## Related issues
-

### DIFF
--- a/v2/emailpassword/serverless/with-aws-lambda/authorizer.mdx
+++ b/v2/emailpassword/serverless/with-aws-lambda/authorizer.mdx
@@ -1,0 +1,113 @@
+---
+id: authorizer
+title: Lambda Authorizer
+hide_title: true
+---
+
+<!-- COPY DOCS -->
+<!-- ./session/serverless/with-aws-lambda/authorizer.mdx -->
+
+# Lambda Authorizer
+
+You can use a lambda similar to [step 5](./session-verification) as an Authorizer in API Gateways. This will enable you to use SuperTokens in a lambda to authorize requests to other integrations (e.g., AppSync).
+
+## 1) Add configurations and dependencies
+
+Follow [step 1](./frontend) and [step 2](./backend-config).
+
+## 2) Create a new file `index.js`
+Use the code below as the handler for the lambda. Remember that whenever we want to use any functions from the `supertokens-node` lib, we have to call the `supertokens.init` function at the top of that serverless function file. We can then use `getSession()` to get the session.
+
+```jsx title="index.js"
+let supertokens = require("supertokens-node");
+let { wrapRequest, wrapResponse } = require("supertokens-node/framework/awsLambda");
+let Session = require("supertokens-node/recipe/session");
+
+let { getBackendConfig } = require("./config");
+
+supertokens.init(getBackendConfig());
+
+exports.handler = async function(event, context) {
+    const wrappedResponse = wrapResponse(event);
+    try {
+        const session = await Session.getSession(wrapRequest(event), wrappedResponse, { sessionRequired: false });
+        if (session) {
+            return generateAllow(session.getUserId(), event.methodArn, {
+                setCookie: event.supertokens.response.cookies.join(', '),
+            });
+        }  else {
+            // You can also return explicit deny with additional context, but you should configure these as 401s so the SDK can handle them.
+            // return generateDeny(undefined, event.methodArn);
+            throw new Error("Unauthorized");
+        }
+    } catch(ex) {
+        throw new Error("Unauthorized");
+    }
+}
+     
+// Help function to generate an IAM policy
+const generatePolicy = function(principalId, effect, resource, context) {
+    // Required output:
+    const authResponse = {};
+    authResponse.principalId = principalId;
+    if (effect && resource) {
+        const policyDocument = {};
+        policyDocument.Version = '2012-10-17'; // default version
+        policyDocument.Statement = [];
+
+        const statementOne = {};
+        statementOne.Action = 'execute-api:Invoke'; // default action
+        statementOne.Effect = effect;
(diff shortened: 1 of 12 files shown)


v2/emailpassword/serverless/with-aws-lambda/authorizer.mdx
```diff
@@ -0,0 +1,113 @@
...
+let Session = require("supertokens-node/recipe/session");
+
+let { getBackendConfig } = require("./config");
+
+supertokens.init(getBackendConfig());
+
+exports.handler = async function(event, context) {
+    const wrappedResponse = wrapResponse(event);
+    try {
+        const session = await Session.getSession(wrapRequest(event), wrappedResponse, { sessionRequired: false });
```
rishabhpoddar, line 33: I believe that wrapping the response manually is not needed. The `getSession` does that automatically so you can call `Session.getSession(event, event, { sessionRequired: false });` and it should work.

v2/emailpassword/serverless/with-aws-lambda/authorizer.mdx
```diff
@@ -0,0 +1,113 @@
...
+        if (session) {
+            return generateAllow(session.getUserId(), event.methodArn, {
+                setCookie: event.supertokens.response.cookies.join(', '),
+            });
+        }  else {
+            // You can also return explicit deny with additional context, but you should configure these as 401s so the SDK can handle them.
```
rishabhpoddar, line 39: Change to "so the frontend SDK can handle them"? (Adding the work frontend to the comment)

v2/emailpassword/serverless/with-aws-lambda/authorizer.mdx
```diff
@@ -0,0 +1,113 @@
...
+            // return generateDeny(undefined, event.methodArn);
+            throw new Error("Unauthorized");
+        }
+    } catch(ex) {
+        throw new Error("Unauthorized");
```
rishabhpoddar, line 44: You should check for the type of the error. If it's `TRY_REFRESH_TOKEN` then send `throw new Error("Unauthorized");`. Otherwise it's a 500 error.

v2/emailpassword/serverless/with-aws-lambda/authorizer.mdx
```diff
@@ -0,0 +1,113 @@
...
+    try {
+        const session = await Session.getSession(wrapRequest(event), wrappedResponse, { sessionRequired: false });
+        if (session) {
+            return generateAllow(session.getUserId(), event.methodArn, {
+                setCookie: event.supertokens.response.cookies.join(', '),
+            });
+        }  else {
+            // You can also return explicit deny with additional context, but you should configure these as 401s so the SDK can handle them.
+            // return generateDeny(undefined, event.methodArn);
+            throw new Error("Unauthorized");
```
rishabhpoddar, line 41: In case session does not exist, it means the user is calling this endpoint without a session. Should we not throw unauthorised here and pass the request to the next integration (AppSync) without the userId header? This way, if there are some operations in graphql that don't require auth, they will still succeed.

v2/emailpassword/serverless/with-aws-lambda/authorizer.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+supertokens.init(getBackendConfig());
+
+exports.handler = async function(event, context) {
+    const wrappedResponse = wrapResponse(event);
+    try {
+        const session = await Session.getSession(wrapRequest(event), wrappedResponse, { sessionRequired: false });
+        if (session) {
+            return generateAllow(session.getUserId(), event.methodArn, {
+                setCookie: event.supertokens.response.cookies.join(', '),
```
rishabhpoddar, line 36: Add a comment explaining why setCookie is needed. Something like "Sometimes `getSession` adds cookies to the response, so we want to capture those and send them to the client."

v2/emailpassword/serverless/with-aws-lambda/authorizer.mdx
```diff
@@ -0,0 +1,113 @@
...
+title: Lambda Authorizer
+hide_title: true
+---
+
+<!-- COPY DOCS -->
+<!-- ./session/serverless/with-aws-lambda/authorizer.mdx -->
+
+# Lambda Authorizer
+
+You can use a lambda similar to [step 5](./session-verification) as an Authorizer in API Gateways. This will enable you to use SuperTokens in a lambda to authorize requests to other integrations (e.g., AppSync).
```
rishabhpoddar, line 12: Maybe you should add somewhere something like "This authorisor will do session verification and if successful, will add the userId of the logged in user to the request object as the `x-user-id` header. If no session exists, this header won't exist either.

v2/thirdparty/serverless/with-aws-lambda/appsync-integration.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+A Lambda Authorizer configured like in the [Authorizer guide](./authorizer) can help integrate Supertokens with an AppSync service.  
+
+## 1) Set up the AppSync service
+
+Set up the AppSync service with an API key authorization. For more details, please see the [AWS documentation](https://docs.aws.amazon.com/appsync/latest/devguide/quickstart.html).
+
+## 2) Configure the API Gateway with the authorizer
+
+Follow the [Authorizer guide](./authorizer) to set up the API Gateway with the `/auth`, and `/graphql` resources set up.
```
rishabhpoddar, line 20: Specify what should one do with the `/auth` gateway (point it to to a lambda function that contains the auth APIs shown in step 3)

rishabhpoddar (changes requested):

Issue #129, opened 2021-09-26
Python sdk documentation


bhumilsarvaiya:

rishabhpoddar:
What have you thought about the underscore vs camel case issue?

PR #141, opened 2021-10-11
quick setup doc update for python (fastapi)

bhumilsarvaiya:
## Summary of change
add quick setup instructions for python sdk

## Related issues
- #129


rishabhpoddar:
- double comma issue in connection_uri
- Use # instead of "//" in CoreInjector component

PR #160, opened 2021-11-22
Use yarn add instead of install

jacobhq:
## Summary of change
Use `yarn add` instead of `yarn install` for commands.

## Checklist


rishabhpoddar:
Thanks @jacobhq

rishabhpoddar:
There are other places too where we had `yarn install`. I will replace those with `yarn add` as well.

PR #209, opened 2022-01-24
Update mysql.mdx

scherbatsky-jr:
Added data type to device_id_hash while creating passwordless_codes table


rishabhpoddar:
Thanks for this fix. You will need to make this fix in [this file](https://github.com/supertokens/docs/blob/master/v2/community/database-setup/mysql.mdx) and in [this file](https://github.com/supertokens/docs/blob/master/v2/community/database-setup/postgresql.mdx) and then run `npm run start` inside v2 dir which will make these changes in all the recipe folders that copy from this file.

rishabhpoddar:
Thanks @scherbatsky-jr

PR #220, opened 2022-02-07
docs: add docs for custom translations

porcellus:
## Summary of change
Adding docs about adding custom translations

## Related issues
- https://github.com/supertokens/supertokens-auth-react/issues/233

## Checklist
- [x] Algolia search needs to be updated? (If there is a new sub docs project, then yes)
- [x] Sitemap needs to be updated? (If there is a new sub docs project, then yes)
- [x] Changes required to the demo apps corresponding to the docs?

### DIFF
--- a/v2/emailpassword/common-customizations/translations.mdx
+++ b/v2/emailpassword/common-customizations/translations.mdx
@@ -1,0 +1,177 @@
+---
+id: translations
+title: Translating the UI
+hide_title: true
+---
+
+<!-- COPY DOCS -->
+<!-- ./passwordless/common-customizations/translations.mdx -->
+
+import FrontendSDKTabs from "/src/components/tabs/FrontendSDKTabs";
+import TabItem from '@theme/TabItem';
+import NpmOrScriptTabs from "/src/components/tabs/NpmOrScriptTabs"
+
+# Translating the UI
+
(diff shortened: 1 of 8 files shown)


rishabhpoddar:
# Feedback
- [ ] It's not clear what "Library integration" means. Consider changing the title.
- [ ] Point 2 and the last para in "Library integration" don't make sense to me..
- [ ] Need a way to tell the user how to get the list of all existing translation keys.
- [ ] Using `currentLanguageCookieScope: "http://localhost"` for cookie scope is incorrect. Please see [this](https://supertokens.com/docs/thirdpartyemailpassword/common-customizations/sessions/share-sessions-across-sub-domains) for how to set cookie scope (in session context, but it's similar here too)
- [ ] It would be useful to provide a code snippet example for "Translations in component overrides" section.

rishabhpoddar:
# Feedback
- [ ] Change the title of "Translating the UI" to "Language Translation for the UI" everywhere.
- [ ] "After initialization, you can load translations by calling the" -> this implies that you have to call the `loadTranslation` function. Maybe change how this is communicated?
- [ ] Typo in "If it's missing also missing from the default store"

rishabhpoddar (approved):

PR #237, opened 2022-02-28
Updated react contribution guide

porcellus:
## Summary of change
Updated contribution guide to reflect the feature component rework

## Related issues
- supertokens/supertokens-auth-react#386

## Checklist
- [x] Algolia search needs to be updated? (If there is a new sub docs project, then yes)
- [x] Sitemap needs to be updated? (If there is a new sub docs project, then yes)
- [x] Changes required to the demo apps corresponding to the docs?


v2/contribute/code-style/react.mdx
```diff
@@ -6,11 +6,46 @@ hide_title: true
...
+## Feature and Theme components
+
+Complex functionality should be implemented in feature components and theme components should only have local, simple, and strictly UI-related states. Some examples:
+
+### Feature component state:
+- Information that is created in one step of the process and used in another. I.e., `loginAttemptInfo` in passwordless sign in/up
+- General error displayed in the sign-in screen but set in the form submit method
+
+### Theme component state:
+- Current value of each form field
```
rishabhpoddar, line 29: One could even have the form fields as part of the feature component state? Why won't we add this as a state in there?

v2/contribute/code-style/react.mdx
```diff
@@ -6,11 +6,46 @@ hide_title: true
...
+- Time left until another retry is allowed
+
+### `useReducer`
+
+You should use a reducer to manage the state of a feature component if:
+- The state is complex (i.e.: more than a single error and/or status field)
+- The feature component functionality will be extracted into a custom hook and reused elsewhere.
+- If other similar feature components need to use a reducer. It's better to be consistent. (e.g.: sign in/up feature components of different auth recipes)
```
rishabhpoddar, line 37: These two points don't make sense to me. Maybe add more details here?

v2/contribute/code-style/react.mdx
```diff
@@ -6,11 +6,46 @@ hide_title: true
...
+
+If you used a reducer to manage the state you should:
+- Move both state and action types into `types.ts` of the recipe.
+- The "root" theme component should receive them as `featureState` and `dispatch`. Further down the component tree, they are usually event handlers and individual state variables.
+- Check out other components for naming conventions, keeping in mind that it can be renamed if it's imported elsewhere
```
rishabhpoddar, line 42: I don't understand "keeping in mind that it can be renamed if it's imported elsewhere"

v2/contribute/code-style/react.mdx
```diff
@@ -6,11 +6,46 @@ hide_title: true
...
+
+### Reusable feature components
+
+If the logic of the feature component is reused (i.e.: the corresponding "root" theme component is used in multiple places), it should be extracted into custom hooks. In most cases, these should be `useFeatureReducer` and `useChildProps`.
```
rishabhpoddar, line 46: If the root theme component is used somewhere else too (like emailpassword's root theme component is used in thirdpartyemailpassword?), then how does adding these custom hook in emailpassword feature component help thirdpartyemailpassword?

v2/contribute/code-style/react.mdx
```diff
@@ -6,11 +6,46 @@ hide_title: true
...
+
+Keep in mind, that some of these custom hooks may be called without the recipe (e.g.: `useChildProps` of the sign-in/up feature components in all auth recipes). In this case, it should:
+- still work (returning undefined is fine)
+- call the same hooks, to not violate the rules of hooks
+- be typed in a way so that the return value is defined if the recipe is also defined.
```
rishabhpoddar, line 51: I don't quite understand this. You will need to add more details:
- What exactly is the purpose of `useChildProps`?
- If this function is defined in a feature component, how can it be called without a recipe?

rishabhpoddar (changes requested):

v2/contribute/code-style/react.mdx
porcellus: The list of form fields (labels, validators, etc.) comes from the feature component (it would usually be part of useChildProps).
We could also elevate the form state (current value, validation status, etc) up to the feature component level, but it's not necessary: 
- we don't use them outside of the `FormBase` component
- we don't want to control or update them from the outside
- they are "temporary": we don't want to persist/restore them after refreshes or navigation.

v2/contribute/code-style/react.mdx
porcellus: That means you don't need to name `useChildProps` as `useEmailPasswordSignInUpChildProps`, since inside the module, the naming is clear, and outside we'd rename them.

v2/contribute/code-style/react.mdx
porcellus: The main responsibility of the feature component is to provide the business logic behind theme components, they are basically the container component of a single theme component ("root"). 
If we want to reuse the root theme component (or a significant portion of it) we will want to also reuse the business logic implementation, which will be easier if they are implemented in custom hooks. This has a few advantages:
- It's strongly typed: the way we used to do this through cloning the child components and overwriting their props was basically passing anys around.
- It's easier to compose: it's much easier to implement interactions between feature components (more precisely these reused bits of business logic) i.e.: getting a single error variable that we can display instead of both subforms handling/showing their own instance of general error.
- It's a much more standard/recommended way to solve this issue.

v2/contribute/code-style/react.mdx
porcellus: In order to not violate the rules of hooks, these can be called without the recipe in "combined" feature components. 
E.g.: in thirdpartyemailpassword sign-in/up with disabled emailpassword or thirdparty recipes. We still need to call these hooks but we don't actually have a recipe instance.

v2/contribute/code-style/react.mdx
rishabhpoddar: okay

v2/contribute/code-style/react.mdx
rishabhpoddar: okay

rishabhpoddar: Consider adding this to the guide's MD file too.

v2/contribute/code-style/react.mdx
rishabhpoddar: I see. okay

v2/contribute/recipe-design/frontend.mdx
```diff
@@ -20,6 +20,48 @@ hide_title: true
 - 9) User interaction with a recipe should be isolated to that recipe, even if that recipe uses several other sub recipes. For example, if a user initialises recipeX (which uses recipeY and recipeZ underneath), the user should not have to do `recipeY.something` or `recipeZ.something`. Therefore, recipeX must re-export useful functions / components that are provided by recipeY and recipeZ.
 - 10) When implementing Recipe interfaces, make sure that if a function uses another function in the interface, it will call the user's overrided verion of that function (if the user has overriden it). See https://github.com/supertokens/supertokens-node/issues/199
 
+## Feature and Theme components
```
rishabhpoddar, line 23:
```suggestion
## React SDK Feature and Theme components
```

rishabhpoddar (changes requested):

PR #245, opened 2022-03-12
fix typo

zvictor:
minor typo fix 😊

### DIFF
--- a/v2/emailpassword/common-customizations/sessions/with-jwt/about.mdx
+++ b/v2/emailpassword/common-customizations/sessions/with-jwt/about.mdx
@@ -27,7 +27,7 @@ In our session flow, we issue an access and a refresh token. Our access token is
 
 However, there are few similarities:
 - Both, our access token, and a JWT can be verified in a stateless manner. Therefore, session verifications are very quick (< 1 MS)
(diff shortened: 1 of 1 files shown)


rishabhpoddar:
hey @zvictor the docs you modified is actually automatically copied from this page: https://github.com/supertokens/docs/blob/master/v2/session/common-customizations/sessions/with-jwt/about.mdx

So you will need to fix the typo from the above page instead, and that will get propagated to the other recipe docs during the build process.

zvictor:
I can do that, no problem.

I ended up here after clicking in the "edit page" link in the docs website. Shouldn't that link point to the right file then?

rishabhpoddar:
@zvictor it actually does link to the right page from Docusauru's point of view. But we have made a custom plugin which allows docs to be copied across recipes. This is something that docusaurus' link to github is not aware of and hence doesn't link to the true source.

That being said, we will investigate if we can make it link to the true source.

zvictor:
> This is something that docusaurus' link to github is not aware of and hence doesn't link to the true source. That being said, we will investigate if we can make it link to the true source.

Yes, that's what I meant 🙂
It should save you time in the future when others do the same mistake I just did.

PR #252, opened 2022-03-23
Docs redesign: Question component

st-jay:
## Summary of change
- Updated the question component
- Updated the `snippetConfigForm/formQuestion` component used in passwordless docs

### Design references
- Design: https://zpl.io/adQE8yK
- Other updates
  - Updated the icons in left nav ( e.g. icons for Netlify, Redwood, etc. )
    - `height: 16px`
    - `width: auto`
    - Space between the link's text and icon should be 8px. The link's text has a `padding-right: 16px`, so added `margin-left: -8px` to icon.
  - Removed `margin-top` for first image that appears inside a tabs container.
  - The `You can refresh the page to undo your selection` message could overlap the answers in the Question component. If the answers overlap the message, the method in `question/index.tsx`, `updateAnswerBoxPaddingIfAnswersOverflow` add extra space to `padding-bottom` of the container which prevents this overlap. It is not needed in the `snippetConfigForm/formQuestion` component as it does not have the message. 

## Related issues
- https://github.com/supertokens/main-website/issues/576
- #257 

## Checklist

- [x] Design review


v2/src/components/question/index.tsx
```diff
@@ -109,43 +71,27 @@ export function Question(props: PropsWithChildren<{
...
+        );
     }
 }
 
 type AnswerProps = {
     title: string,
-    onClick?: () => void
-}
+    onClick?: () => void,
+    isLast: boolean
```
nkshah2, line 81: What is this used for?

nkshah2:
Screenshot 2022-03-23 at 11 48 11 AM

Screenshot 2022-03-23 at 11 48 04 AM

I think the default question component added some spacing above and below it

v2/src/components/question/index.tsx
```diff
@@ -1,86 +1,75 @@
...
+            const availableWidth = answerBoxWidth - messageWidth - 36;
+
+            let totalAnswersWidth = 0;
+            for (const answer of answersInsideContainer) {
+                const width = answer.getBoundingClientRect().width;
+                totalAnswersWidth = totalAnswersWidth + width + 24;
+            }
+            setExtraPaddingToAnswerBox(totalAnswersWidth > availableWidth);
+        }
+    };
```
nkshah2, line 37: Why was this required? I thought this was just a design change

st-jay: it was needed to be conditional, if the answers overlap the message, then only we add extra padding, otherwise we do not add extra padding

nkshah2: What does overlap mean here?

nkshah2: Also `const messageWidth = 227;` where did you get this number from? and how will this work on mobile?

st-jay: on iPad Pro, the message `Refresh the page to undo your selection` overlaps the answer buttons.

I got the `const messageWidth = 227;` from the width of `Refresh the page to undo your selection`. Though I realise now it would be better to get the width on execution, otherwise we would need to update this value any time we change the style or content of that message. 

image

nkshah2: Wouldnt it make more sense to give the text a fixed width based on its parent width instead? (and the same for the buttons container too)

nkshah2: Also this happens if I open the preview on an actual iPad Pro

IMG_64B76BD3DEF6-1

Issue #254, opened 2022-03-23
Incorrect error message is shown in appInfoForm for Netlify Functions


st-jay:
## Bug
- Visit https://supertokens.com/docs/passwordless/serverless/with-netlify/frontend
- In the appInfoForm, input `/.netlify/functions/` for `apiBasePath` and toggle the `I am using Netlify Serverless Functions` checkbox to true.
- Fill in all other fields correctly.
- On submit, the form gives this error - _*apiBasePath should begin with '/.netlify/functions/' when using Netlify Serverless Functions._
- This error is incorrect and should be updated.

PR #264, opened 2022-03-27
Bug fix for AppInfoForm and updated error message for invalid apiBasePath when using Netlify functions

### LINKED ISSUE #263: AppInfoForm bug: apiDomain and websiteDomain do not ignore case for Localhost
## Problem
- For the `apiDomain` and `websiteDomain` fields inside `AppInfoForm`, if we insert values non-lowercase values of "localhost" ( like `Localhost`, `LOCALHOST`, etc ), it gives an 'Invalid domain' error.
- As it considers non-lowercase `localhost` values invalid, if we use `Localhost` with different ports for `apiDomain` and `websiteDomain`, it results to `""`. This as a result, gives a false error that `apiBasePath` cannot be equal to, be a prefix of or be a prefix to `websiteBasePath`.

## Solution
- Convert the domains to lower case before normalising the domain. As we use JS' `URL` object in the domain normalization process, the output domain is already in lower case, so converting the input to lowercase too doesn't change the original output.

st-jay:
## Related issues
- #263 
- #254 

## Summary of changes
- fixed #263 - Convert the domain to lower case using `String.toLowerCase()` method when calling the normalization class `NormalisedURLDomain`.
- #254: Updated the error message we show when using Netlify Functions and the apiBasePath is not prefixed by `/.netlify/functions/` or does not contain any path after that.
  - Updated message: **"apiBasePath must be prefixed with \`/.netlify/functions\`, for example to use \`/auth\` set the apiBasePath to \`/.netlify/functions/auth\`"**

## Checklist

No todos remaining

### DIFF
--- a/v2/src/components/appInfoForm/index.tsx
+++ b/v2/src/components/appInfoForm/index.tsx
@@ -671,11 +671,7 @@ export default class AppInfoForm extends React.PureComponent<PropsWithChildren<P
                     ) {
                         // if the netlify api route checkbox is set to true
                         // the api base path can only start with `/.netlify/functions`
-                        validationErrors.apiBasePath = "apiBasePath must be prefixed with `/.netlify/functions`, for example to use `/auth` set the apiBasePath to `/.netlify/functions/auth`";
-                    } else if (netlifyApiRouteUsed && validApiBasePath === "/.netlify/functions/") {
-                        // if the netlify api route is `/.netlify/functions/` without any scope
-                        // we show this error, as functions at `/.netlify/functions/` route aren't possible
-                        validationErrors.apiBasePath = "apiBasePath should be of the format '/.netlify/functions/*' when using Netlify Serverless Functions.";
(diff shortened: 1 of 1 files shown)


v2/src/components/appInfoForm/index.tsx
```diff
@@ -667,11 +671,7 @@ export default class AppInfoForm extends React.PureComponent<PropsWithChildren<P
                     ) {
                         // if the netlify api route checkbox is set to true
                         // the api base path can only start with `/.netlify/functions`
-                        validationErrors.apiBasePath = "apiBasePath should begin with '/.netlify/functions/' when using Netlify Serverless Functions."
-                    } else if (netlifyApiRouteUsed && validApiBasePath === "/.netlify/functions/") {
```
st-jay, line 671: @nkshah2 I've removed this condition as the `validApiBasePath` will never be equal to `/.netlify/functions/` as the path normalization function we use removes the trailing `/`

nkshah2 (approved):

Issue #267, opened 2022-03-30
[BUG] TypeError due to question component inside passwordless docs on Safari 12


st-jay:
## Problem
- Open the passwordless docs ( https://supertokens.com/docs/passwordless/quick-setup/frontend ) on Safari 12
- choose a value from the question component ( one that gives you the options to choose `Phone number`, `email`, `phone number or email` )
- It gives the following error and the page breaks-
```
TypeError: t.replaceAll is not a function. (In 't.replaceAll("^{form_"+r.id+"}",o)', 't.replaceAll' is undefined)
```

PR #271, opened 2022-03-31
Remove usage of replaceAll from question component in passwordless docs

st-jay:
## Summary of change
- The question component in passwordless docs uses `replaceAll` method in strings, which is not available in Safari 12.
- We replace this `replaceAll(stringWhichIsReplaced, stringThatReplaces)` by `.split(stringWhichIsReplaced).join(stringThatReplaces)`. This method is used in the `AppInfoForm` and works on Safari 12.

### Demo
- [iPad Air iOS v12.4](https://drive.google.com/file/d/1xSkouLU4Vfe25JORA0mZgDKIYX1cb-JD/view?usp=sharing)

## Related issues
- #267 

## Checklist

No todos remaining

### DIFF
--- a/v2/src/components/snippetConfigForm/index.tsx
+++ b/v2/src/components/snippetConfigForm/index.tsx
@@ -11,6 +11,20 @@ import { recursiveMap } from "../utils";
  *
  */
 
+// polyfill for `replaceAll` as it is not supported in Safari v12
+// @ts-ignore
+if (String.prototype.replaceAll === undefined || String.prototype.replaceAll === null) {
+    function escapeRegExp (str: string) {
+        return str.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"); // $& means the whole matched string
+    }
+
+    // @ts-ignore
+    String.prototype.replaceAll = function (find: string, replace: string) {
+        let target = this;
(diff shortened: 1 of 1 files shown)


v2/src/components/snippetConfigForm/index.tsx
```diff
@@ -119,13 +119,13 @@ export default class SnippetConfigForm<T extends keyof any> extends React.PureCo
               if (typeof c === "string") {
                 for (const [ind, question] of this.props.questions.entries()) {
                   const value = this.state.answers[ind];
-                  c = c.replaceAll(`^{form_${question.id}}`, value);
+                  c = c.split(`^{form_${question.id}}`).join(value);
```
rishabhpoddar, line 122: We actually have a pollyfill for replace all that we use: https://github.com/supertokens/docs/blob/master/v2/src/plugins/copyDocsAndCodeTypeChecking.js#L11

You should use that, or copy that function and use that?

rishabhpoddar (changes requested):

v2/src/components/snippetConfigForm/index.tsx
st-jay: updated

rishabhpoddar (approved):

PR #295, opened 2022-04-24
complaince changes when adding new RDS cluster

bhumilsarvaiya:
## Summary of change
- adding steps for complaince while adding new RDS cluster
- how to remove pending postgresql connections when trying to remove a saas app

### DIFF
--- a/v2/contribute/checklists/saas/add-new-region.md
+++ b/v2/contribute/checklists/saas/add-new-region.md
@@ -55,6 +55,68 @@ hide_title: true
 
 - RDS
     - Create based on other regions rds
+    - For each db instance withing the cluster, add tags:
+        - VantaContainsUserData: true
+        - VantaDescription: db cluster to store development/production containers data
+        - VantaOwner: rishabh@supertokens.com
+    - Add CloudWatch Alarms
+        - FreeableMemory Alarm
+            - Name: db-{{RDS-cluster-name}} FreeableMemory
+            - Type: Metric alarm
+            - Namespace: AWS/RDS
+            - Metric name: FreeableMemory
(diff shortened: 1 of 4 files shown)


v2/contribute/checklists/saas/add-new-region.md
```diff
@@ -55,6 +55,68 @@ hide_title: true
 
 - RDS
```
rishabhpoddar, line 56:
```suggestion
- Add PostgreSQL RDS
```

v2/contribute/checklists/saas/mysql-rds.md
```diff
@@ -1,3 +1,8 @@
+---
+id: mysql-rds
+title: Launching new Mysql RDS Cluster
+hide_title: true
+---
```
rishabhpoddar, line 5: - How come this is added in this PR but the file existed before? Was this forgotten to be added before?
- Also, describe why we are using a MySQL RDS when we already have a postgresql rds section in the new-region.md page.

rishabhpoddar (changes requested):

v2/contribute/checklists/saas/mysql-rds.md
bhumilsarvaiya: > How come this is added in this PR but the file existed before? Was this forgotten to be added before?

Yes, it was forgotten to be added

> Also, describe why we are using a MySQL RDS when we already have a postgresql rds section in the new-region.md page.

There was mysql connection pool error where the old connections were not removed because the mysql procedure was not set. I made that change in DB and you asked me to put the changes in this file for reference so that if we ever move/create new mysql rds instance, that issue doesn't occur

rishabhpoddar (approved):

PR #319, opened 2022-05-06
Created plugin to add `rel="nofollow"` to external links in docs

st-jay:
## Summary of change
- Created a rehype plugin that modifies the nodes of docs that are links.
- It checks if the link is an external link and adds `rel="nofollow noopener noreferrer"` ( I included `noopener noreferrer` because that was already being added before this plugin. )
- The reason why I added a rehype plugin instead of a remark plugin, is that after the remark plugins make changes to nodes, docusaurus was adding `_target="blank" rel="noopener noreferrer"` to the links and discarding the `nofollow` added by the remark plugin.

## Related issues
- #317 

## Checklist

- [x] Testing

### DIFF
--- a/v2/src/plugins/addNofollowToExternalLinks.js
+++ b/v2/src/plugins/addNofollowToExternalLinks.js
@@ -1,0 +1,84 @@
+const url = require("url");
+
+module.exports = () => {
+  /**
+   * Check if the URL is external or not
+   * @param {string} href the URL that is to be checked
+   * @returns boolean true if the URL is external and false otherwise
+   */
+  function isUrlExternal (href) {
+    try {
+      const hrefFromURL = new url.URL(href);
+      const host = hrefFromURL.host;
+      const protocol = hrefFromURL.protocol;
+
+      // no nofollow needed for email or phone number links
+      if (protocol === "mailto:" || protocol === "tel:") {
+        return false;
+      }
+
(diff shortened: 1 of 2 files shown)


v2/src/plugins/addNofollowToExternalLinks.js
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Add `nofollow` to the element passed if it is an external link and recursively do the same to the element's children
+   * @param {Node} data element node from the document's tree
+   * @returns updated element with `nofollow` in external links
+   */
+  function addNofollowToExternalLinks(data) {
+    try {
+      if (shouldAddNofollowToLink(data)) {
+        data.properties = Object.assign({}, data.properties, {
+          rel: "nofollow noreferrer noopener"
```
nkshah2, line 58: For links going to one of our Github repos we still want to keep referrer information (we use Github analytics information as well). 

Doing this for all links is not a good idea

nkshah2 (changes requested):

v2/src/plugins/addNofollowToExternalLinks.js
st-jay: `noopener noreferrer` was already being added by docusaurus before, so I have kept it here as well

nkshah2 (approved):

PR #349, opened 2022-06-21
Implements angular code tabs

jscyo:
## Summary of change
Adds angular docs

## Related issues
- none

## Checklist

- if override for the render function is not required in the demo app, the demo app will need to be updated
- if ngOnChange is not required then the demo app will need to be updated


jscyo:
## Feedback
- [ ] change mentions of auth routes from `auth` to `/auth`
- [ ] remove the point about optimizing load times with code splitting
- [ ] clearly mention that the `supertokens-auth-react` SDK is responsible for login UI and Auth flows
- [ ] move the example app to the end of the first section and make it important
- [ ] mention where the react file should be created
- [ ] have proper names for the code snippets, full file paths
- [ ] do you need ngOnChange?
- [ ] why is there override next to the render function
- [ ] Add another section in the Step 2 init section about initializing `supertokens-website`

rishabhpoddar (changes requested):

PR #352, opened 2022-06-22
chore: Facebook doc update

sattvikc:
## Checklist


rishabhpoddar:
Need to make this change in all places

PR #379, opened 2022-07-08
fix: fixed snippets for golang

sattvikc:
## Summary of change
Fixed snippets for golang breaking changes

## Checklist

### DIFF
--- a/v2/emailpassword/common-customizations/sessions/update-jwt-payload.mdx
+++ b/v2/emailpassword/common-customizations/sessions/update-jwt-payload.mdx
@@ -414,10 +414,14 @@ func main() {
 		currAccessTokenPayload["newKey"] = "newValue"
 
 		//highlight-next-line
-		err = session.UpdateAccessTokenPayload(handle, currAccessTokenPayload)
+		updated, err := session.UpdateAccessTokenPayload(handle, currAccessTokenPayload)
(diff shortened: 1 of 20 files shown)


v2/emailpassword/common-customizations/sessions/update-jwt-payload.mdx
```diff
@@ -414,11 +414,15 @@ func main() {
...
-		err = session.UpdateAccessTokenPayload(handle, currAccessTokenPayload)
+		updated, err := session.UpdateAccessTokenPayload(handle, currAccessTokenPayload)
 		if err != nil {
 			// TODO: handle error
 			return
 		}
+		if !updated {
+			// Session does not exist
+			continue
+		}
```
rishabhpoddar, line 425: You don't really need this since there is nothing at the end of this loop anyway. Instead, in this loop, you get `sessionInfo` which should be a pointer now, so you need to check if that's nil, and if it is, then you need to call continue.

rishabhpoddar: Similar applies to other changes in the PR

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #427, opened 2022-08-10
Session claims docs

rishabhpoddar:
## TODO
- [x] new collapsable component for docs -> https://app.zeplin.io/project/5aaa0de98d835c7b25637f2e/screen/62f507765761426ab1724cdf (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) (@sulha199 will do)
- [x] Add email verification to passwordless and third party recipe
- [x] In email verification enable section, talk about enabling on the backend + frontend. For frontend, we also need to give code snippets for angular and vanilla JS.
- [x] In email verification, we need a page on protecting routes with email verification on the frontend / backend (similar to how we have for user roles)
- [x] Need to describe OPTIONAL mode for email verification.
- [x] "Changing the website path for the email verification UI (optional)" section needs to change to use the correct method + `GetEmailVerificationURL` is not longer a function in golang and python as well
- [x] "Changing the website path for reset password UI (optional)" needs to change to use the right function for nodejs and also `GetResetPasswordURL` is no longer used in golang + python
- [x] Changes to framework integration docs? (Important to check AWS lambda authoriser)
- [x] Changes to 2fa and phone password docs
- [x] Replace use of updateAccessTokenPayload with mergeIntoAccessTokenPayload
- [x] Add emailderlivery to thirdparty
- [x] update embed sign up / in component docs to use `doesSessionExist` from sessionContext and if session does exist, wrap that child components in sessionAuth (talk to @rishabhpoddar if this doesn't make sense)
- [x] golang
- [x] python
- [x] AWS lambda authoriser changes? -> handling INVALID_CLAIMS and not just the TRY_REFRESH_TOKEN error type.
- [x] NextJS server side rendering situation when it throws INVALID_CLAIMS?


v2/emailpassword/common-customizations/email-verification/changing-style.mdx
```diff
@@ -0,0 +1,212 @@
...
+        Session.init()
+    ]
+});
+```
+</TabItem>
+</FrontendSDKTabs>
+
+The above will result in: 
+
+<img alt="Prebuilt form with custom submit button" width="500px" src="/img/emailpassword/signup-centered-blue-button.png" />
```
rishabhpoddar, line 67: this is an incorrect image link

PR #434, opened 2022-08-19
update style for collapsible

sulha199:
## Summary of change
Update collapsible styles according to https://app.zeplin.io/project/5aaa0de98d835c7b25637f2e/screen/62f507765761426ab1724cdf

- update `summary` style
- update `details` border
- slight animation when expand/collapse

local test url: http://localhost:3000/docs/change_me/collapsible-element

Note: the change_me example's summary uses 20px font. if it use the normal size, then the summary will have 48px height like the design.


v2/src/css/custom.css
```diff
@@ -519,3 +519,48 @@ html[data-theme="dark"] .markdown blockquote > *:last-child {
...
+  font-size: 16px;
+  cursor: pointer;
+}
+
+details > summary::before {
+  content: "";
+  position: relative;
+  box-sizing: border-box;
+  display: inline-block;
+  background: url(/img/icon-plus.svg) center;
```
nkshah2, line 546: The svg appears different on different devices, use a png instead

v2/src/css/custom.css
```diff
@@ -519,3 +519,48 @@ html[data-theme="dark"] .markdown blockquote > *:last-child {
     bottom: 144px !important;
   }
 }
+
+/* Styles for collapsible elements */
+details {
+  position: relative;
+  transition: 0.3s, 0s border;
```
nkshah2, line 526: The transition feels really strange especially since the text inside has no transition. It just pops in and then the rest of the detail component transitions in and out.

We should just remove the transition all together (both the detail and the ::before)

v2/src/css/custom.css
```diff
@@ -519,3 +519,48 @@ html[data-theme="dark"] .markdown blockquote > *:last-child {
...
+  margin: 6px 0;
```
nkshah2, line 527: Why is this margin here?

v2/src/css/custom.css
```diff
@@ -519,3 +519,48 @@ html[data-theme="dark"] .markdown blockquote > *:last-child {
...
+}
+
+details > summary {
+  list-style-type: none;
+  position: relative;  
+  display: flex;
+  align-items: center;
+  background-color: var(--ifm-pre-background);  
+  padding: 11px 20px;
```
nkshah2, line 536: Should be 14px 20px based on the designs

v2/src/css/custom.css
```diff
@@ -519,3 +519,48 @@ html[data-theme="dark"] .markdown blockquote > *:last-child {
...
+  cursor: pointer;
+}
+
+details > summary::before {
+  content: "";
+  position: relative;
+  box-sizing: border-box;
+  display: inline-block;
+  background: url(/img/icon-plus.svg) center;
+  background-size: cover;
```
nkshah2, line 547: Refrain from using cover, use contain or stretch (contain preferably)

nkshah2 (changes requested):

v2/src/css/custom.css
sulha199: to prevent two collapsible to stick to each other vertically.

So, I added margin, and compensate padding

v2/src/css/custom.css
sulha199: its removed

v2/src/css/custom.css
sulha199: updated

v2/src/css/custom.css
sulha199: updated

v2/src/css/custom.css
sulha199: updated

v2/src/css/custom.css
```diff
@@ -519,3 +519,45 @@ html[data-theme="dark"] .markdown blockquote > *:last-child {
...
+  background-size: contain;
+  width: 20px;
+  height: 20px;
+  border-radius: 50%;
+  margin-right: 12px;
+}
+
+details[open] {
+  padding-bottom: 20px;
+  border-bottom: 1px solid var(--ifm-toc-border-color);;
```
nkshah2, line 554: The detail component in the design has a border top and bottom in both open and closed states (Its very light so it is easy to miss in the design)

v2/src/css/custom.css
```diff
@@ -519,3 +519,45 @@ html[data-theme="dark"] .markdown blockquote > *:last-child {
     bottom: 144px !important;
   }
 }
+
+/* Styles for collapsible elements */
+details {
```
nkshah2, line 524: Lets change the `change_me/collapsible-element` page to add a default detail component as well, the current ones force font size to 20 px so hard to review the output

```mdx
<details>
<summary>Default State</summary>
This is the default state of the `detail` component which uses font size 16px
</details>
```

nkshah2 (changes requested):

v2/src/css/custom.css
sulha199: its updated

nkshah2 (approved):

PR #451, opened 2022-09-01
fix: avoid run isMobile on SSG

sulha199:
## Summary of change
only run `isMobile` on browser and not during SSG

## Related issues
- #450 

## Checklist

NA

### DIFF
--- a/v2/src/theme/NavbarItem/RecipeLabelItem.js
+++ b/v2/src/theme/NavbarItem/RecipeLabelItem.js
@@ -1,18 +1,15 @@
 import React from 'react';
-import { isMobile } from 'react-device-detect';
+import { BrowserView, isMobile } from 'react-device-detect';
 import GuidesLink from "../../components/guidesLink";
 
 export default function RecipeLabelItem(props) {
-    if (isMobile) {
-        return null;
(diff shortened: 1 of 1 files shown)


v2/src/theme/NavbarItem/RecipeLabelItem.js
```diff
@@ -1,18 +1,15 @@
 import React from 'react';
-import { isMobile } from 'react-device-detect';
+import { BrowserView, isMobile } from 'react-device-detect';
```
nkshah2, line 2: https://user-images.githubusercontent.com/18233774/188368841-cf8f3905-ae93-4e8d-b568-6ca63696dbd5.MP4

Still seeing the issue (Video from iOS safari)

nkshah2 (changes requested):

v2/src/theme/NavbarItem/RecipeLabelItem.js
sulha199: so its a SSG case. now its fixed by using docusaurus's `BrowserOnly`

nkshah2 (approved):

PR #529, opened 2022-11-28
ADR: last week discussions

porcellus:
## Summary of change
- Adding ADRs for last weeks discussions
- Removed an outdated note 

## Related issues
- 

## Checklist


v2/contribute/decisions/passwordless/0003-claim-redirection-on-fe.mdx
```diff
@@ -0,0 +1,61 @@
...
+## Considered Options
+
+* Recommend creating wrapper components
+* Add a callback that returns a react component to render on failure
+* Add callbacks (onFailure, onSuccess) that returns a string (or undefined) we can redirect to
+* Add a separate component that handles these checks/mapping
+
+## Decision Outcome
+
+Add callbacks (onFailure, onSuccess) that returns a string (or undefined) we can redirect to.
```
rishabhpoddar, line 31: would the string be just a path? or even full URLs? Pros / cons of that? (I ask this cause for example, in getRedirect function, we only allow return of path)

v2/contribute/decisions/passwordless/0004-combined-doesSessionExist-and-validateClaims.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+## Decision Outcome
+
+Add a combined function (`hasAccess`):
+- new function to web-js that combines doesSessionExists and validateClaims
+- no redirection
+- Returns: `{ hasAccess: true } | {hasAccess: false, invalidClaims: [], sessionExists: boolean}`
+
+Reasons:
+- A separate `validateClaims` is just too easy to forget.
```
rishabhpoddar, line 33: As a side note, this imples that emailverification.init also takes the mode now

v2/contribute/decisions/session/0022-handshake-after-jwt-access-token.mdx
```diff
@@ -0,0 +1,39 @@
...
+* Keep the handshake
+* Remove handshake
+
+## Decision Outcome
+
+We can drop the handshake:
+- Anti-csrf mode can be defined on the backend
+- Access token blacklisting will be moved to the backend
+- Validities are already sent in the create and refresh core API responses
+- The core JWKs endpoint can return the necessary public keys
```
rishabhpoddar, line 35: Would dropping handshake also work for if JWT key rotation is enabled such that:
- Only new keys are added periodically
- New keys are added periodically and older keys are auto removed.

Is there any edge case you can think of in which the SDK will fail to query the JWKs endpoint? Or maybe the user does not want the endpoint to exist (which is possible if they disable that API), so then the SDK will have to use the handshake way as a fallback?

rishabhpoddar (changes requested):

v2/contribute/decisions/passwordless/0004-combined-doesSessionExist-and-validateClaims.mdx
porcellus: If we want to do this kind of redirection in web-js, I think this should also handle redirection if the session doesn't exist. I'll try and come up with something about that.
Either that or not handle redirection in web-js (which means we still wouldn't need the mode input)

v2/contribute/decisions/session/0022-handshake-after-jwt-access-token.mdx
porcellus: - Looking at how JWK client libs seem to handle caching (i.e.: https://github.com/auth0/node-jwks-rsa/blob/master/EXAMPLES.md#caching), dropping the handshake would work in both cases.
- This should still be queried from the core, so the backend failing to communicate with it would be a problem. I don't think we should disable this endpoint there. This could have the same auth checks as other core endpoints, we could make that part configurable, but I don't see a reason to disable this endpoint as it only returns public information.

rishabhpoddar: Making it so that user's can't disable this endpoint is going against our architecture. Which i don't like. That being said, the point of just querying the core instead makes sense, however, the format in which the core returns the public keys is very different and not directly consumable by jwks libraries. So we might need to change that maybe?

porcellus: I was talking about a (possibly new) core jwks endpoint, that conforms to the standard. I'm unsure how much we differ from the standard besides the location. 
The core should still be the one signing the access tokens, so we have to query the core to fetch the public keys.
The only other option is signing in the backend SDKs (but I don't think we should). Then we wouldn't need to retrieve any keys to verify the tokens because it's right there in the backend SDK memory.

v2/contribute/decisions/session/0021-handshake-after-jwt-access-token.mdx
```diff
@@ -32,7 +32,9 @@ We can drop the handshake:
 - Anti-csrf mode can be defined on the backend
 - Access token blacklisting will be moved to the backend
 - Validities are already sent in the create and refresh core API responses
-- The core JWKs endpoint can return the necessary public keys
+- The core JWKs endpoint can return the necessary public keys. 
+- Since these JWTs (the access token) should still be signed in the Core, the validation should fetch the public keys from the Core as well, which should also conform to the standard.
+- These public keys can be mirrored in the jwks endpoint exposed by the backend SDK if necessary (e.g.: other services/FE want to validate the JWT signature), but this can be disabled.
```
rishabhpoddar, line 37:
```suggestion
- These public keys can be mirrored in the jwks endpoint exposed by the backend SDK if necessary (e.g.: other services/FE want to validate the JWT signature), but this can be disabled by the user if they want to.
```

v2/contribute/decisions/session/0021-handshake-after-jwt-access-token.mdx
```diff
@@ -32,7 +32,9 @@ We can drop the handshake:
 - Anti-csrf mode can be defined on the backend
 - Access token blacklisting will be moved to the backend
 - Validities are already sent in the create and refresh core API responses
-- The core JWKs endpoint can return the necessary public keys
+- The core JWKs endpoint can return the necessary public keys. 
+- Since these JWTs (the access token) should still be signed in the Core, the validation should fetch the public keys from the Core as well, which should also conform to the standard.
```
rishabhpoddar, line 36:
```suggestion
- Since these JWTs (the access token) should still be signed in the Core, the validation should fetch the public keys from the Core as well, which should also conform to the standard. This implies that this route on the core should not expect the apiKey.
```

v2/contribute/decisions/sessionclaims/0003-claim-redirection-targets-on-fe.mdx
```diff
@@ -0,0 +1,34 @@
...
+We have decided to use configurable callbacks to redirect on claim validation success/failure (see [here](./0001)). These functions determine where we redirect the user by returning a string. We have to decide if this string is a full URL or just a path
+
+## Considered Options
+
+* Path
+* Full URL
+
+## Decision Outcome
+
+The return value should be taken interpreted as a full URL (with the option to return just the path for local redirections):
```
rishabhpoddar, line 27:
```suggestion
The return value should be interpreted as a full URL (with the option to return just the path for local redirections):
```

v2/contribute/decisions/sessionclaims/0003-claim-redirection-targets-on-fe.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+- This doesn't matter if there is no session sharing between subdomains. This is the case for most sites.
+- It makes sense for `onFailure` to redirect to the website domain in many cases (e.g: email verification check failing)
+- There are cases, where `onFailure` doesn't point to the website domain, e.g.: the access denied page is specific to the app/subdomain we are currently on
+- `onSuccess` is even more likely to redirect to a domain that doesn't match the websitedomain (e.g: continue browsing after email verification/2FA on websitedomain)
+- Even if we could make onFailure redirect only to the website domain, doing the same to `onSuccess` would block too many usecases (or cause awkward double redirections).
```
rishabhpoddar, line 33: Should we also allow onRedirectionUrl callbacks to return full domains in this case? Cause right now, they are only allowed to return a path.

v2/contribute/decisions/passwordless/0004-combined-doesSessionExist-and-validateClaims.mdx
rishabhpoddar: No no. We don't handle redirection in web-js. But the mode would tell the SDK to add the email verification claim checks automatically when the user calls `hasAccess` - so that in case email is not verified and mode is REQUIRED, then `hasAccess` will return false - as expected.

v2/contribute/decisions/session/0022-handshake-after-jwt-access-token.mdx
rishabhpoddar: Making the core URL conform with the standard makes sense.

v2/contribute/decisions/passwordless/0004-combined-doesSessionExist-and-validateClaims.mdx
porcellus: Oh, ok, I thought I misunderstood something when we last talked about this. I agree; we shouldn't do redirections there.
Right now, the email verification validator is always added to the validators (even in the optional mode in auth-react). The difference between optional and required is the redirection only, but the check is done, and the validator appears in the `invalidClaims` array for both. I'll add a new ADR as clarification.

v2/contribute/decisions/sessionclaims/0003-claim-redirection-targets-on-fe.mdx
porcellus: `getRedirectUrl` can return full URLs, the `with-one-login-many-subdomains` example uses this.

v2/contribute/decisions/passwordless/0004-combined-doesSessionExist-and-validateClaims.mdx
rishabhpoddar: Right. But we need to change that now? Cause if mode is optional, the hasAccess should return true - so we should not add the validators if mode is optional?

rishabhpoddar (approved):

PR #548, opened 2022-12-08
Thirdparty ADRs 7-9

sattvikc:
## Summary of change

ADRs 7 to 9

## Related issues
- Link to issue1 here
- Link to issue1 here

## Checklist

- [ ] Item1
- [ ] Item2


v2/contribute/decisions/thirdparty/0007-check-for-enabled-recipes.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+### Recipes should not check
+
+Recipes functions/APIs don't check for this flag, and, all the statically configured recipes will function as defined, even if the user has disabled them in the core.
+
+<ArgumentCon> Inconsistent SDK behaviour with the user configuration. For example, a user might be able to create an emailpassword account and use the application, which a particular tenant intends to disable, just because it was statically declared for use with a different tenant. </ArgumentCon>
+
+### Recipes should check
+
+The recipe functions/APIs check if the recipe is enabled in the core and then continue it's execution
```
rishabhpoddar, line 44: I think only the APIs should check for this and not the functions - we should still allow the dev to call the functions manually for a tenant even if they have not enabled that recipe for that tenant. The reason for this is that it provides more flexibility and also has no security issues. The reason for checking in API is cause those are public facing and there pose a security risk.

v2/contribute/decisions/thirdparty/0008-default-tenant-creation-in-core.mdx
```diff
@@ -0,0 +1,42 @@
...
+import ArgumentCon from "/src/components/decisionLogs/ArgumentCon"
+import ArgumentNeut from "/src/components/decisionLogs/ArgumentNeut"
+
+# Creation of default tenant in core
+
+<DecisionHeader status="proposed" lastUpdate="2022-10-25" created="2022-10-25" deciders={["rishabhpoddar", "porcellus"]} proposedBy={["porcellus"]} />
+
+## Context and Problem Statement
+
+Since the recipe functions/APIs query the core if the recipe is enabled in core, what is the desired behaviour for the default tenant
```
rishabhpoddar, line 18: huh? This sentence is not properly written.

v2/contribute/decisions/thirdparty/0008-default-tenant-creation-in-core.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+## Considered Options
+
+* Do not create a default tenant in core, assume all recipes are enabled if config not found
+* Create a default tenant in core with all recipes (emailpassword, passwordless and thirdparty) enabled
```
rishabhpoddar, line 23:
```suggestion
* Do not create a default tenant in core, assume all recipes are enabled if config not found
* Create a default tenant in core with all recipes (emailpassword, passwordless and thirdparty) enabled
* Do not create a default tenant in core and assume no recipes are enabled by default.
```

v2/contribute/decisions/thirdparty/0008-default-tenant-creation-in-core.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+
+## Decision Outcome
+
+Chosen option: **Create a default tenant in core with all recipes enabled**, because
+
+* While adding thirdparty providers to the core, there is no ambiguity if the recipe should be enabled at that time
```
rishabhpoddar, line 30: huh? what does this even mean?

v2/contribute/decisions/thirdparty/0008-default-tenant-creation-in-core.mdx
```diff
@@ -0,0 +1,42 @@
...
+* Recipe functions/APIs can check if the recipe is enabled without having to handle the corner cases
+
+## Pros and Cons of the Options
+
+### Do not create a default tenant in core, assume all recipes are enabled if config not found
+
+<ArgumentCon> Might unintentionally enable all recipes for an unknown tenant, for which config won't be found </ArgumentCon>
```
rishabhpoddar, line 37: I don't understand this point at all.

v2/contribute/decisions/thirdparty/0008-default-tenant-creation-in-core.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+### Create a default tenant in core with all recipes enabled
+
+<ArgumentNeut> Safe default for the default tenant </ArgumentNeut>
+<ArgumentPro> No ambiguity when adding thirdparty providers, as the enabled flag is already set to true </ArgumentPro>
```
rishabhpoddar, line 42: how is this pro point even related to this ADR? 

The reason for creating a default tenant in the core is so that by default, for single tenant all the recipes will work if they are initialised by the dev - which is the expected dev experience. If we don't make a default tenant, we will have to have edge case checks everywhere to enable recipes if no tenant exists and if the user is using single tenancy.

v2/contribute/decisions/thirdparty/0009-config-management-functions-when-recipe-is-disabled.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+Chosen option: Should be allowed, because
+
+* it is more flexible
+
+## Pros and Cons of the Options
+
+### Should not be allowed
+
+<ArgumentCon> Ambiguity when the user enables a recipe, adds the config and then disables it. Should already configured values be retained or destroyed? </ArgumentCon>
```
rishabhpoddar, line 36: they don't have to be destroyed per say.. just not be returned. This would have the same effect as what we have now - just with a worse developer experience.

v2/contribute/decisions/thirdparty/0009-config-management-functions-when-recipe-is-disabled.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+### Should be allowed
+
+<ArgumentPro> Recipes can be enabled and disabled at ease, without the need for re-configuration </ArgumentPro>
```
rishabhpoddar, line 40: need for re config would not be there if we do not destroy the config.. the only reason we allow management of the config is cause of a better developer experience and nothing else.

rishabhpoddar (changes requested):

v2/contribute/decisions/thirdparty/0007-check-for-enabled-recipes.mdx
```diff
@@ -35,12 +35,15 @@ Chosen option: **Recipes should check**, because
 
 ### Recipes should not check
 
-Recipes functions/APIs don't check for this flag, and, all the statically configured recipes will function as defined, even if the user has disabled them in the core.
```
rishabhpoddar: make it very clear that recipe functions don't check and only APIs check and write down why we decided this.

v2/contribute/decisions/thirdparty/0008-initial-state-of-default-tenant-in-core.mdx
```diff
@@ -0,0 +1,50 @@
...
+import ArgumentCon from "/src/components/decisionLogs/ArgumentCon"
+import ArgumentNeut from "/src/components/decisionLogs/ArgumentNeut"
+
+# Initial state of default tenant in core
+
+<DecisionHeader status="proposed" lastUpdate="2022-10-25" created="2022-10-25" deciders={["rishabhpoddar", "porcellus"]} proposedBy={["porcellus"]} />
+
+## Context and Problem Statement
+
+Since the recipe APIs query the core if the recipe is enabled in core,
```
rishabhpoddar, line 18:
```suggestion
Since the backend SDK's recipe APIs query to check the core about if the recipe is enabled for the input (or defaul) tenant,
```

v2/contribute/decisions/thirdparty/0009-config-management-functions-when-recipe-is-disabled.mdx
```diff
@@ -27,14 +27,14 @@ In the context of thirdparty, should management of dynamic provider list be allo
...
 
-* it is more flexible
+* it is a better developer experience (less restricting)
 
 ## Pros and Cons of the Options
 
 ### Should not be allowed
 
-<ArgumentCon> Ambiguity when the user enables a recipe, adds the config and then disables it. Should already configured values be retained or destroyed? </ArgumentCon>
+<ArgumentCon> User will be forced to enable a recipe before adding it's configs </ArgumentCon>
```
rishabhpoddar, line 36:
```suggestion
<ArgumentCon> User will be forced to enable a recipe before adding its configs </ArgumentCon>
```

v2/contribute/decisions/thirdparty/0009-config-management-functions-when-recipe-is-disabled.mdx
```diff
@@ -27,14 +27,14 @@ In the context of thirdparty, should management of dynamic provider list be allo
...
 
 ### Should be allowed
 
-<ArgumentPro> Recipes can be enabled and disabled at ease, without the need for re-configuration </ArgumentPro>
+<ArgumentPro> Better user experience as he can add/update all the configs and then enable the recipe </ArgumentPro>
```
rishabhpoddar, line 40:
```suggestion
<ArgumentPro> Better user experience as she can add/update all the configs and then enable the recipe </ArgumentPro>
```

rishabhpoddar (changes requested):

v2/contribute/decisions/thirdparty/0007-check-for-enabled-recipes.mdx
```diff
@@ -21,8 +21,14 @@ Should each of the recipe check if the recipe is enabled in the core?
...
 
-* Recipes should not check
+* Recipe should not check
+  * API Should not check
+  * Functions should not check
 * Recipes should check
+  * API should check
+  * Functions should not check
+
+Reason why we don't check in the functions because it allows the user to use the functions even if that recipe is not enabled for that tenant, and is not exposed to pose security risk. However, API check would need to be considered as it is publicly exposed.
```
rishabhpoddar, line 31:
```suggestion
Reason why we don't check in the functions is because it allows the user to use the functions even if that recipe is not enabled for that tenant. This makes the dev experience is better. Furthermore, it does not pose any security risk since these recipe functions are not exposed publicly. However in the case of APIs, we need to check if the recipe is enabled for the tenant because they are publicly exposed and want to avoid a situation in which a tenant can sign up, for example, with emailpassword even though that recipe is not enabled for them in the core.
```

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #562, opened 2022-12-25
adr: moving override config to separate component's prop

alisher-aituarov:
## Summary of change
Add discussion on moving components override config out of init call

## Related issues
- https://github.com/supertokens/supertokens-auth-react/issues/616
## Checklist

### DIFF
--- a/v2/contribute/decisions/reactsdk/0001-move-override-config-to-separate-component-prop.mdx
+++ b/v2/contribute/decisions/reactsdk/0001-move-override-config-to-separate-component-prop.mdx
@@ -15,7 +15,7 @@ import ArgumentNeut from "/src/components/decisionLogs/ArgumentNeut"
 
 ## Context and Problem Statement
 
-We may want to separate override config from supertokens-auth-react init, so user does have to call `init` both for `web-js` and `auth-react`. See related [issue](https://github.com/supertokens/supertokens-auth-react/issues/616)
+We may want to separate override config from supertokens-auth-react init, so user does not have to call `init` both for `web-js` and `auth-react`. See related [issue](https://github.com/supertokens/supertokens-auth-react/issues/616)
 
 ## Considered Options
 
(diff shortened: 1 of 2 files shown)


v2/contribute/decisions/reactsdk/0001-move-override-config-to-separate-component-prop.mdx
```diff
@@ -0,0 +1,46 @@
...
+import ArgumentCon from "/src/components/decisionLogs/ArgumentCon"
+import ArgumentNeut from "/src/components/decisionLogs/ArgumentNeut"
+
+# Move override config to separate component's prop
+
+<DecisionHeader status="proposed" lastUpdate="2022-12-21" created="2022-12-21" deciders={["rishabhpoddar", "porcellus"]} proposedBy={["rishabhpoddar"]} />
+
+## Context and Problem Statement
+
+We may want to separate override config from supertokens-auth-react init, so user does have to call `init` both for `web-js` and `auth-react`. See related [issue](https://github.com/supertokens/supertokens-auth-react/issues/616)
```
rishabhpoddar, line 18:
```suggestion
We may want to separate override config from supertokens-auth-react init, so user does not have to call `init` both for `web-js` and `auth-react`. See related [issue](https://github.com/supertokens/supertokens-auth-react/issues/616)
```

rishabhpoddar, line 18: This reasoning is incorrect. We want to extract the override config for components only (and not functions) so that the auth-react SDK can be initialised without having to depend on react.

rishabhpoddar: Note that the functions override remains in the auth-react init as it is right now.

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #581, opened 2023-01-22
Adding ADR about handling front-token after flattening the access token

porcellus:
## Summary of change
Adding ADR about handling front-token after flattening the access token

## Related issues
- 

## Checklist

### DIFF
--- a/v2/contribute/decisions/session/0015-jwt-compatibility.mdx
+++ b/v2/contribute/decisions/session/0015-jwt-compatibility.mdx
@@ -30,5 +30,6 @@ Once we make the entire access token into a JWT:
 - Since the access token can now be used instead of the a separate JWT embedded in the access token, it is more useful to send to third-party APIs where we are not intercepting request.
     - These will check the access token based on the JWKs endpoint exposed by the BE, but we do not need to worry about key rotation here, since the signing keys stay in the output of the jwks endpoint long after the access token expires.
     - Calling `getAccessToken` will cause a refresh to happen if the locally stored access token is expired, so refreshing should happen automatically in most cases (if the access token doesn't expire during the request) even if we are not intercepting requests to the domain.
+    - Claim checking needs to be done manually in this case (e.g.: an API Gateway will not check if the email is verified or not). This should be documented/communicated somewhere.
 
 Until then: the auth mode and jwt options are entirely independent, the jwt can be accessed through the access token payload in both auth modes.
--- a/v2/contribute/decisions/session/0015-jwt-compatibility.mdx
+++ b/v2/contribute/decisions/session/0015-jwt-compatibility.mdx
@@ -27,5 +27,8 @@ We want to make our access token into a valid JWT. So, in the future the jwt set
 Once we make the entire access token into a JWT:
 - Add deprecation warning with a migration link and set `exposeAccessTokenToFrontendInCookieBasedAuth` to true. See [here](./0022)
 - Add a function to the FE to access the entire access token as a string.
+- Since the access token can now be used instead of the a separate JWT embedded in the access token, it is more useful to send to third-party APIs where we are not intercepting request.
+    - These will check the access token based on the JWKs endpoint exposed by the BE, but we do not need to worry about key rotation here, since the signing keys stay in the output of the jwks endpoint long after the access token expires.
+    - Calling `getAccessToken` will cause a refresh to happen if the locally stored access token is expired, so refreshing should happen automatically in most cases (if the access token doesn't expire during the request) even if we are not intercepting requests to the domain.
 
 Until then: the auth mode and jwt options are entirely independent, the jwt can be accessed through the access token payload in both auth modes.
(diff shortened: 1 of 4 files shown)


v2/contribute/decisions/session/0015-jwt-compatibility.mdx
```diff
@@ -27,5 +27,8 @@ We want to make our access token into a valid JWT. So, in the future the jwt set
 Once we make the entire access token into a JWT:
 - Add deprecation warning with a migration link and set `exposeAccessTokenToFrontendInCookieBasedAuth` to true. See [here](./0022)
 - Add a function to the FE to access the entire access token as a string.
+- Since the access token can now be used instead of the a separate JWT embedded in the access token, it is more useful to send to third-party APIs where we are not intercepting request.
+    - These will check the access token based on the JWKs endpoint exposed by the BE, but we do not need to worry about key rotation here, since the signing keys stay in the output of the jwks endpoint long after the access token expires.
+    - Calling `getAccessToken` will cause a refresh to happen if the locally stored access token is expired, so refreshing should happen automatically in most cases (if the access token doesn't expire during the request) even if we are not intercepting requests to the domain.
```
rishabhpoddar, line 32: there is an issue with this (which we may not need to solve, but worth pointing out) - if the user has email verification as required, and if they use some APi gateway for session verification (instead of using verifySession or getSession from our SDK), then it's on them to check the email verification claim in the session before granting access.

This is not communicated in the docs anywhere though, and it's not obvious that this needs to be done. I assume other auth providers solve this problem cause they don't even return the id token back to the client application if the email is not verified.

v2/contribute/decisions/session/0031-internal-prop-handling-in-flat-access-tokens.mdx
```diff
@@ -0,0 +1,62 @@
...
+- Expose all claims
+
+## Decision Outcome
+
+We should hide internal-only claims
+
+- We don't want to confuse our users
+- They are never used in applications
+
+We decided to hide: `refreshTokenHash1`, `parentRefreshTokenHash1`, because they do not make sense for the application and are never used as a part of CDI or FDI.
```
rishabhpoddar, line 39: One issue with this is that it causes inconsistencies when reading the access token payload on the backend and frontend. The backend will show these "hidden" values, but the frontend wont.

Therefore, if there is no security risk in exposing these to the frontend (which i don't think there is), then we should just expose them.

v2/contribute/decisions/session/0031-internal-prop-handling-in-flat-access-tokens.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+<ArgumentPro> Matches the current behaviour </ArgumentPro>
+<ArgumentPro> Hides all claims that could be confusing </ArgumentPro>
+<ArgumentCon> People have been looking for the userId claim </ArgumentCon>
+<ArgumentCon> We are hiding otherwise standard claims </ArgumentCon>
+
+### Expose all claims 
+
+<ArgumentPro> Front-token is very straght-forward </ArgumentPro>
+<ArgumentCon> Changes the current behaviour/interface </ArgumentCon>
```
rishabhpoddar, line 59: why would there by changes to the "current behaviour/interface"?

v2/contribute/decisions/session/0031-internal-prop-handling-in-flat-access-tokens.mdx
```diff
@@ -0,0 +1,62 @@
...
+<ArgumentPro> Exposes the standard claims (people have tried to use them through getAccessTokenPayload before) </ArgumentPro>
+<ArgumentCon> Could be confusing, we'd need to document each internal claim (and/or move them into a separate `st_internal` claim) </ArgumentCon>
```
rishabhpoddar, line 61: not really.. we don't need to document this. If someone asks, we can just answer somewhere. Not a big deal.

rishabhpoddar (changes requested):

v2/contribute/decisions/session/0031-internal-prop-handling-in-flat-access-tokens.mdx
porcellus: New things show up in the access token payload. This change can affect apps that loop through the props in the payload.

v2/contribute/decisions/session/0031-internal-prop-handling-in-flat-access-tokens.mdx
porcellus: We could "hide" them from the `getAccessTokenPayload` on the backend; I assumed that both implementations would have to return the same thing.
Anyway, if you think these extra props are not too confusing, I'm happy with just exposing them all.

rishabhpoddar: Yea, i don't like the idea of hiding in general. Especially since the access token is a JWT and when someone sees it on a jwt inspector site, they will see those extra claims anyway

rishabhpoddar (approved):

PR #584, opened 2023-01-24
adr: add adr for splitting recipe routes from recipe init

porcellus:
## Summary of change

Add adr for splitting recipe routes from recipe init:
We want to enable apps to initialize supertokens-auth-react without bundling react. This is useful for example in angular apps, where we should be able to use a single init call in the main route (and use auth-react throughout the app), while only bundling react to the module handling auth.

## Related issues
- 

## Checklist

### DIFF
--- a/v2/contribute/decisions/reactsdk/0002-split-recipe-route-handlers-into-separate-comps.mdx
+++ b/v2/contribute/decisions/reactsdk/0002-split-recipe-route-handlers-into-separate-comps.mdx
@@ -1,0 +1,142 @@
+---
+id: "0002"
+title: Split recipe-specific route handlers into separate components
+hide_title: true
+---
+
+import DecisionHeader from "/src/components/decisionLogs/DecisionHeader";
+import ArgumentPro from "/src/components/decisionLogs/ArgumentPro";
+import ArgumentCon from "/src/components/decisionLogs/ArgumentCon";
+import ArgumentNeut from "/src/components/decisionLogs/ArgumentNeut";
+
+# Split recipe-specific route handlers into separate components
+
+<DecisionHeader
+  status="proposed"
+  lastUpdate="2023-01-24"
+  created="2023-01-24"
+  deciders={["rishabhpoddar", "porcellus"]}
+  proposedBy={["porcellus"]}
+/>
+
+## Context and Problem Statement
+
+We want to enable apps to initialize supertokens-auth-react without bundling react. This is useful for example in angular apps, where we should be able to use a single init call in the main route (and use auth-react throughout the app), while only bundling react to the module handling auth.
+
+## Considered Options
+
+- Add recipeIds into input of getRoutingComponent
+- Add separate components for the pre-built UI of each recipe
+
+## Decision Outcome
+
+Chosen option: Add separate components for the pre-built UI of each recipe, because
+
+- Interface fits well into react
+- Works with most bundlers
+
+Notes:
+
+- The new components will be called `RecipeNamePreBuiltUIRoutes`
+- The new components should take `router` as a required and `componentOverrides` as an optional prop.
+- `canHandleRoute` will be moved into recipe index files
+
+## Pros and Cons of the Options
+
+### Add recipeIds into input of getRoutingComponent
+
+<ArgumentPro> Could work with really smart bundlers </ArgumentPro>
+<ArgumentPro> Less interface change </ArgumentPro>
+<ArgumentCon> The recipes need to be listed in two places </ArgumentCon>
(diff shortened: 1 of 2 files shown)


v2/contribute/decisions/reactsdk/0002-split-recipe-route-handlers-into-separate-comps.mdx
```diff
@@ -0,0 +1,142 @@
...
+    <SuperTokensWrapper>
+      <div className="App">
+        <Router>
+          <Routes>
+            <EmailPasswordPreBuiltUIRoutes router={require("react-router-dom")} />
+            <EmailVerificationPreBuiltUIRoutes
+              router={require("react-router-dom")}
+              componentOverrides={{
+                EmailVerificationSendVerifyEmail_Override: () => <OtpScreen />,
+              }}
```
rishabhpoddar, line 105: we should not have componentOverride here

v2/contribute/decisions/reactsdk/0002-split-recipe-route-handlers-into-separate-comps.mdx
```diff
@@ -0,0 +1,142 @@
...
+```tsx
+function App() {
+  return (
+    <SuperTokensWrapper>
+      <div className="App">
+        <Router>
+          <Routes>
+            <EmailPasswordPreBuiltUIRoutes router={require("react-router-dom")} />
+            <EmailVerificationPreBuiltUIRoutes
+              router={require("react-router-dom")}
```
rishabhpoddar, line 102: need to investigate effect on disableDefaultUI

rishabhpoddar (changes requested):

v2/contribute/decisions/reactsdk/0002-split-recipe-route-handlers-into-separate-comps.mdx
```diff
@@ -94,27 +96,27 @@ With react-router-dom:
...
-      </div>
+      <EmailVerificationComponentsOverrideProvider
+        components={{
+          EmailVerificationSendVerifyEmail_Override: () => <OtpScreen />,
+        }}>
+        <div className="App">
+          <Router>
+            <Routes>
+              <EmailPasswordPreBuiltUIRoutes router={require("react-router-dom")} />
+              <EmailVerificationPreBuiltUIRoutes router={require("react-router-dom")} />
```
rishabhpoddar, line 107: can this be used to render these on a dynamic route? For example for different languages

rishabhpoddar (changes requested):

v2/contribute/decisions/reactsdk/0002-split-recipe-route-handlers-into-separate-comps.mdx
rishabhpoddar: @porcellus missed this? Can you add more details about this - if we can / should do this via these components or some other method. Another way of doing this would be to make the websiteBasePath a regex  (which can be done later on as well)

porcellus: No, I just haven't had time to address this.
This method could be used to do this, but it'll take a bit further work. Right now, this would only match paths starting with websitebasepath.

rishabhpoddar: So let's keep it the way it is. We can address the issue of dynamic websiteBasepath later on

v2/contribute/decisions/reactsdk/0002-split-recipe-route-handlers-into-separate-comps.mdx
```diff
@@ -0,0 +1,144 @@
...
+class App extends React.Component {
+  render() {
+    // We could handle looping through these for the user but I don't think this is an issue.
+    if (EmailPasswordPreBuiltUI.canHandleRoute()) {
+      return EmailPasswordPreBuiltUI.getRoutingComponent();
+    }
+
+    if (EmailVerificationPreBuiltUI.canHandleRoute()) {
+      return EmailVerificationPreBuiltUI.getRoutingComponent({
+        EmailVerificationSendVerifyEmail_Override: () => <OtpScreen />,
```
rishabhpoddar, line 137: Would this have component override? Cause i thought we would be using `EmailVerificationComponentsOverrideProvider`, in which case, how would component override work here exactly?

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #664, opened 2023-05-23
Oauth2 ADRs

porcellus:
## Summary of change

## Related issues
- Link to issue1 here
- Link to issue1 here

## Checklist

- [ ] Item1
- [ ] Item2


v2/contribute/decisions/oauth/0008-customizable-token-builders.mdx
```diff
@@ -0,0 +1,56 @@
...
+    - These functions will get all information we have about the current flow (requested scopes, user info, etc)
+    - These functions will get the "default" access token payload as a parameter. This will include:
+        - all default claims (e.g.: `sub`)
+        - other claims if we choose to add default implementation for other scopes (e.g.: `supertokens.email` could add the user email and verification status into the id_token)
+    - These functions will NOT be affected by whatever is added to `createNewSession` (in the Session recipe)
+- We will recommend people to add third-party integration (i.e.: hasura) related claims into the access token by overriding `createNewAccessToken`
+    - They fit the resource server role in OAuth2.
+    - This can be done based on the requested/approved scopes in the override
+    - Competitors go with either this or adding custom claims into the id_token.
+    - Both tokens are customizable, so both solutions work. This is mostly about recommendation
```
rishabhpoddar, line 43:
```suggestion
    - Both tokens are customizable, so both solutions work. This is mostly about recommendation
    - Users should use the access token (and not the id token) to query the oauth server as well as their own backend. Their own backend includes third party services like hasura, so it makes sense to recommend people to add custom claims into the access token and not id token.
```

v2/contribute/decisions/oauth/0009-account-linking.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+* Use user id mapping
+* Use account linking
+
+## Decision Outcome
+
+Chosen option: Use account linking if using ST as a third-party provider
+
+* Doesn't require using separate apps (as in multitenancy) for the provider and the consumer.
+
```
rishabhpoddar, line 30: One important point here that we had discussed is that this should continue to work even if the account linking feature flag is switched off - this exception case can be easily done on the core side by not checking for the license key features when doing account linking for this use case.

v2/contribute/decisions/oauth/0010-m2m-admin-access.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+## Considered Options
+
+* Provide separate API for admin keys
+* Recommend access tokens using custom scopes for M2M
+
+## Decision Outcome
+
+Chosen option: Recommend access tokens using custom scopes for M2M
+
```
rishabhpoddar, line 28: is this m2m between client backend and the oauth server? Or client backend to another client backend? Is there any distinction that needs to be made between these two?

rishabhpoddar: I ask cause I assume the option of "Provide separate API for admin keys" doesn't really apply to m2m between client backend to another client backend right?

v2/contribute/decisions/oauth/0010-m2m-admin-access.mdx
```diff
@@ -0,0 +1,46 @@
...
+* Fits the client credential flow of OAuth2 well.
+* Doesn't require a separate/dedicated recipe for M2M.
+
+## Pros and Cons of the Options
```
rishabhpoddar, line 32: related to this, how will can client credential flow be used by our users to allow them to issue API keys to their customers so that their customers can query our user's backend using the API key?

Usually, what i have seen is that our customer would give a client id + secret to their customers, who would then use the client credential flow to get a new access token, which can then be used to query our customer's application backend. In this case, since our access tokens are JWTs, how can our customers revoke them?

Finally, can this client credentials flow be used to just give one single API key to customers and using just that, they can directly access our customer's application's backend apis?

v2/contribute/decisions/oauth/0011-frontend-routes.mdx
```diff
@@ -0,0 +1,48 @@
...
+When the user redirected to the app during the OAuth2 sign-in process, there are a few routes we can choose to provide.
+
+## Considered Options
+
+* Merge functionality into existing auth page
+* Add new route for OAuth2 specific authentication that embeds the sign in/up components
+* Add new route that redirects to default auth page
+
+## Decision Outcome
+
```
rishabhpoddar, line 27: Add that this route also is redirected back to from the auth page post login and then this route will redirect the user back to the client app

v2/contribute/decisions/oauth/0012-refresh-token-rotation.mdx
```diff
@@ -0,0 +1,42 @@
...
+## Considered Options
+
+* No refresh token rotation
+* Always use rotating refresh tokens
+
+## Decision Outcome
+
+Chosen option: Always use rotating refresh tokens
+
+* Fits the newer security best practices
```
rishabhpoddar, line 29: Mention that we will not take care of many of the edge cases related to this and are OK with this.

Should we have a config that can be used to switch off refresh token rotation here? Maybe a core config? I can imagine that there might be users who just want to keep things simple and not care so much about security. Forcing them to also keep track of changing refresh tokens can be a big ask.

v2/contribute/decisions/oauth/0014-not-expiring-tokens.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+In some M2M use-cases the user may require a token to never expire.
+
+## Considered Options
+
+* Recommend long expiration times instead of non-expiring tokens
+* Set the expiration time to null for non-expiring tokens
+
+## Decision Outcome
+
```
rishabhpoddar, line 26: This refers to access token expiry right? Does this also apply to refresh token expiry?

rishabhpoddar: If the access tokens are very login lived and the user wants to revoke them - how will they do it? Is there something in the access token which the user can check in the db to see if it's revoked or not - how can users check this on a code level (for example for our sessions, we have a `checkDatabase` boolean)?

v2/contribute/decisions/oauth/0015-storing-client-secret.mdx
```diff
@@ -0,0 +1,49 @@
...
+## Context and Problem Statement
+
+Some other providers let you see the client secret multiple times.
+
+## Considered Options
+
+* Only store the hash and not support viewing it later
+* Store the secrets as plaintext
+* Store the secrets encrypted
+
```
rishabhpoddar, line 25: Another option is to store the hash and the first few chars of the secret. So users can kind of know which secret belongs to which client if they have many pairs. But i don;t think we need to do this. But worth mentioning

v2/contribute/decisions/oauth/0011-frontend-routes.mdx
```diff
@@ -0,0 +1,48 @@
...
+# Redirect to the standard login route if required
+
+<DecisionHeader status="proposed" lastUpdate="2023-05-11" created="2023-05-11" deciders={["rishabhpoddar", "porcellus"]} proposedBy={["porcellus"]} />
+
+## Context and Problem Statement
+
+When the user redirected to the app during the OAuth2 sign-in process, there are a few routes we can choose to provide.
+
+## Considered Options
+
```
rishabhpoddar, line 21: releated to the frontend is also - how will we store the scopes, redirect_uri etc from the start of the flow till the end? Cause logging in can be a multi step process involving email verification, 2fa etc.. We also need to keep in mind that something like passwordless login will end up creating a session in the other browser (if the link is opened on another browser), so in this case, how will the scopes, redirect uri be transferred?

rishabhpoddar (changes requested):

v2/contribute/decisions/oauth/0002-supported-flows.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+## Considered Options
+
+* All flows
+* Only OAuth2.0 flows (rfc6749)
+* Non-deprecated flows in rfc6749 + PKCE
+
+## Decision Outcome
+
+Chosen option: Non-deprecated flows in rfc6749 + PKCE, because
```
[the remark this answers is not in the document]
porcellus: The refresh token grant flow is in rfc6749 (so included in this choice).
The client id + secret is only required for confidential clients. See here: https://www.rfc-editor.org/rfc/rfc6749#section-6

v2/contribute/decisions/oauth/0004-scope-list.mdx
```diff
@@ -0,0 +1,39 @@
+---
+id: "0004"
+title: Store a global list of scopes that can be enabled per client
+hide_title: true
+---
+
+import DecisionHeader from "/src/components/decisionLogs/DecisionHeader"
```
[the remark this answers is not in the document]
porcellus: We can allow a client to be created without any scopes - they could be added in separate requests.
I think a sensible default is that the `id_token` scope is allowed for clients. In this case we have a sensible default for requesting tokens without specifying scopes, but we do not add any permissions by default.

We should not add all reserved scopes as a default: take the example of a claim that gives access to the phone number/second factor information of the user. This is something we do not want to expose by default.

v2/contribute/decisions/oauth/0004-scope-list.mdx
```diff
@@ -0,0 +1,39 @@
+---
+id: "0004"
+title: Store a global list of scopes that can be enabled per client
+hide_title: true
+---
```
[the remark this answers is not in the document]
porcellus: The standard says:
```
   If the client omits the scope parameter when requesting
   authorization, the authorization server MUST either process the
   request using a pre-defined default value or fail the request
   indicating an invalid scope.
```
Based on this, I suggest defaulting to an `openid` scope. In this case the `id_token` should only contain the default JWT claims (sub, iat, exp, iss, aud) but no other info like email or phone number.

v2/contribute/decisions/oauth/0005-reserved-scopes.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+# Reserve scopes starting with `supertokens.`
+
+<DecisionHeader status="proposed" lastUpdate="2023-05-11" created="2023-05-11" deciders={["rishabhpoddar", "porcellus"]} proposedBy={["porcellus"]} />
+
+## Context and Problem Statement
+
+When creating scopes we can choose what to allow and/or reserve for special use
+
+## Considered Options
```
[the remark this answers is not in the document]
porcellus: That format suggestion is in the title of this page (I'm also adding it to the outcome), but either works for me.

v2/community/database-setup/postgresql.mdx
```diff
@@ -304,6 +304,89 @@ CREATE TABLE IF NOT EXISTS totp_used_codes (
 );
 
 CREATE INDEX totp_used_codes_expiry_time_ms_index ON totp_used_codes(expiry_time_ms);
```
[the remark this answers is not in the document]
porcellus: I'll move them into a separate PR

v2/contribute/decisions/oauth/0007-use-jwt-access-tokens.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+<DecisionHeader status="proposed" lastUpdate="2023-05-11" created="2023-05-11" deciders={["rishabhpoddar", "porcellus"]} proposedBy={["porcellus"]} />
+
+## Context and Problem Statement
+
+When building the OAuth2 access & id_tokens we can make them customizable.
+
+## Considered Options
+
+* Only standard claims
```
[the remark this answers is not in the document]
porcellus: That's what I wanted to add, not sure where/when it got lost.

v2/contribute/decisions/oauth/0014-not-expiring-tokens.mdx
porcellus: - This refers to the access token, but the refresh token expiration is also customizable.
- We expose a token verification endpoint that checks the database. In this case they have to basically handle these as opaque tokens.

v2/contribute/decisions/oauth/0012-refresh-token-rotation.mdx
porcellus: Those users are most likely also using a lib that handles tokens for them, which should also support refresh tokens.

I'm not sure if this should be a core config or an oauth2 client level setting. The core config should be very easy to add later, so I think it's fine if we do not implement it at first.

v2/contribute/decisions/oauth/0011-frontend-routes.mdx
porcellus: 1. I'd say we should store them in sessionstorage (localstorage is also an option):
    - Navigating inside the page (and waiting for an email to be verified for example) works
    - No security issues with someone starting a login process, closing the tab/browser then you opening it and logging into something unintentionally.
    - If the email was verified in another tab/device we can continue the flow after a refresh (we may want to implement some kind of polling for this to not require refreshing)
2. We can't really continue the login process on another device (or at least it would not be standard): 
    - even if we added all info into the magic link (or transfered it to the other device through other means) the state would be missing from the client after the redirection
    - this also matches the general magic link flow we have - the original tab keeps checking for an active session then goes into a "You have been successfully signed in. Please Close this tab" state (which will have to change to check if it needs to do this success redirect)

v2/contribute/decisions/oauth/0010-m2m-admin-access.mdx
porcellus: 1. I don't think we need to make a distinction. The SDK is not exposing special endpoints on the oauth server that is accessible by oauth access tokens
2. Separate API for admin keys can apply to both. It'd have to be a separate recipe for the express purpose of authenticating M2M requests (both on the requester and the requested sides)

v2/contribute/decisions/oauth/0010-m2m-admin-access.mdx
porcellus: 1. They create an access token using the client credential flow - this access token can be handled asn an opaque "API key"
2. We've discussed that flow as well
    - It will be available (this is about what we recommend)
    - We've discussed that issuing a long lived token can be easier to use from the POV of the customer (of our customer)
3. They can be (and in this case should be) handled as opaque tokens. We expose a token validation endpoint for this purpose.
4. Yes - basically the issued access token is the api key.

v2/contribute/decisions/oauth/0004-scope-list.mdx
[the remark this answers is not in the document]
rishabhpoddar: Right. This makes sense. @AdithyaSSup please take a note of this since this will affect the auth code creation endpoint in the core. Also, `openid` is a reserved scope that doesn't start with `supertokens`  - so we need to prevent users from deleting this scope.

v2/contribute/decisions/oauth/0005-reserved-scopes.mdx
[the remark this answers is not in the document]
rishabhpoddar: ok. Let's go ahead with `supertokens.*`. @AdithyaSSup - please take note. Also @porcellus , we should define the list of reserved scopes in this ADR. Here are a few:
- `openid`
- `supertokens.roles` (adds the roles of the user in the access and the id token - or just one of them?)
- `supertokens.profile`(adds the name of the user (if present in the metadata as first_name, last_name keys), email, email verification status, phone number, third party info). Should these be added to the id token and access token or just one of them?

What else?

Also, if the dev does not add `supertokens.profile` scope in the auth code flow, do we still add the user's email and verification status in the id and access token?

v2/contribute/decisions/oauth/0004-scope-list.mdx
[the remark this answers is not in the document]
rishabhpoddar: Makes sense. @AdithyaSSup please take note. We should remove the `addDefaultScopes` boolean from the client creation API. If the input scopes array is empty, we add just the `openid` scope to the client. So we also need to change the flow diagram in this API to not bother with querying the get scopes function.

v2/contribute/decisions/oauth/0010-m2m-admin-access.mdx
rishabhpoddar: > The SDK is not exposing special endpoints on the oauth server that is accessible by oauth access tokens

Shouldn't it? For example to get the profile info of the user with the access token? So if a flow doesn't request openid scope, then how will the client's app get the user's information?

> Separate API for admin keys can apply to both. It'd have to be a separate recipe for the express purpose of authenticating M2M requests (both on the requester and the requested sides)

I don't want to add any new recipe for this - it should all be do-able with oauth 2 recipe ideally

v2/contribute/decisions/oauth/0010-m2m-admin-access.mdx
rishabhpoddar: Understood.

v2/contribute/decisions/oauth/0012-refresh-token-rotation.mdx
rishabhpoddar: Sounds good. Please add this to the ADR, that we decided to purposely not add this core config for now and can do it later if required (as opposed to us just not having thought about it)

v2/contribute/decisions/oauth/0014-not-expiring-tokens.mdx
rishabhpoddar: Sounds good. Please add these points to this ADR. Also mention that the expiry thing applies to refresh tokens as well - there will be a default config in the core for refresh token lifetime, but the sdk can specify a different value.

Also, once an access / refresh token pair is created, and when the refresh token is being used to generate new tokens - those tokens lifetime can't be changed for that session right? 

Also this has implications on the db structure - we will need to store the token's lifetime per token row in the db (since they can be different for different tokens) - @AdithyaSSup please see

v2/contribute/decisions/oauth/0011-frontend-routes.mdx
rishabhpoddar: So this breaks the passwordless login flow that we have - cause when the magic link is clicked and opens in a different browser (not tab), the session is only created in the other browser. The original browser doesn't know about the session creation and is in the same page as "check your email". The way things are right now, passwordless login via magic links won't work for oauth 2.

This probably also explains why all other providers log the user into the original browser and not the new browser (this was a question we were pondering over at the time we were implementing passwordless login).

Ideally, i would not like to change the passwordless magic link flow now (cause that requires work etc..), but instead, we could disallow people from using magic link based login when using in the context of an oauth 2 provider. What do you think?

rishabhpoddar: Also, using session storage may not be the best idea cause it's only per tab. If the user goes through the email verification flow, and opens the link in a new tab (same browser), then the session storage on that tab will not have the right info - breaking the oauth flow.

v2/contribute/decisions/oauth/0004-scope-list.mdx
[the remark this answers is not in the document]
porcellus: I think we do not need to add it to the client in the DB - since this is always enabled for all clients.

v2/contribute/decisions/oauth/0004-scope-list.mdx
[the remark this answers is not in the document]
porcellus: 1. I don't think we even need to add it to the allowed scopes for client table, since this is always allowed and cannot be deleted.
2. No, we don't really need to add it to the scope list if it exists; if the client is not requesting an id_token we should not generate one.

v2/contribute/decisions/oauth/0005-reserved-scopes.mdx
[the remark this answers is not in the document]
porcellus: 1. We should only add these claims to the id token by default. The access token is not supposed to contain this kind of info (but we allow it to be customized)
2. we should not add any info outside of the user ID to the tokens without an appropriate scope: adding potentially unnecessary PII into the tokens by default and making the user add customization to remove it is not great. These things should be opt-in in general and adding a profile scope to the request is as easy as it gets.

v2/contribute/decisions/oauth/0002-supported-flows.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+<DecisionHeader status="proposed" lastUpdate="2023-05-11" created="2023-05-11" deciders={["rishabhpoddar", "porcellus"]} proposedBy={["porcellus"]} />
+
+## Context and Problem Statement
+
+The OAuth2 docs/rfcs contain a lot of different grant types/flows, we need to decide to which to implement. There are a few groups I'd like to consider:
+- OAuth2.0 (the ones defined in rfc6749) grant types: Authorization Code, Implicit, Resource Owner Password Credentials, Client Credentials
+- Refresh token (defined in rfc6749, but not as a grant type)
+- Authorization Code w/ PKCE (not defined in rfc6749)
+- Other extension grant type (e.g.: Device code)
```
AdithyaSSup, line 22: Nit: AFAIK, `device code` grant type is not an extension grant type, rather its one among the grant types.
Also , `refresh token` is not a standalone flow/grant type but rather an extension to other grant types.

porcellus: - refresh token is not a grant type (as it's noted above) and technically not a separate flow either. I can note that as well, the main point here was that we need to decide to support refresh tokens or not.
- I meant extension grant type as described here: https://www.rfc-editor.org/rfc/rfc6749#section-4.5

v2/contribute/decisions/oauth/0010-m2m-admin-access.mdx
porcellus: > Shouldn't it?

We can (and should) expose a profile endpoint that returns whatever we'd add to the id token for an access token. I meant using the M2M access tokens. The profile endpoint does not really make sense using a token without an associated user (created by the client credentials flow).

v2/contribute/decisions/oauth/0014-not-expiring-tokens.mdx
porcellus: I think allowing users to change the lifetime of the refreshed token is OK. Not sure about use-cases, but it's possible that someone would require a session "upgrade" where after some extra validation (e.g.: MFA). Not sure if anybody requires this, but I don't see an issue of making this configurable when calling refresh either.
Keep in mind, that this will not be exposed to the frontend and in most cases will not even be used by the backend (which will fall back to the default set in the core config.)

v2/contribute/decisions/oauth/0003-consent-screen.mdx
```diff
@@ -0,0 +1,57 @@
...
+Chosen option: No built-in consent screen at first, because
+
+* We can get initial support earlier
+* We can ensure future versions will not need migration
+
+What we need store to enable this in the future:
+- enabled scopes (per client):
+    - required anyway
+- auto allowed scopes (per client):
+    - these scopes are allowed ("consented to") by default for every user, meaning
```
AdithyaSSup, line 37: @rishabhpoddar to support this feature we may have to add an additional field (`consent required `)in `client allowed table`..
Should i add it now ?

v2/contribute/decisions/oauth/0004-scope-list.mdx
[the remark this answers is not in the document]
AdithyaSSup: @rishabhpoddar .. As @porcellus suggested, we can have `openid` as an implicit default scope. Then in such a case ,there isn't any need to store/delete, and if the user tries to remove this scope from the required list of scopes , then such a request can be invalidated.

However , I have few questions pertaining to addition of default scope.
- if the user tries add `openid` to the list of required scopes during client creation, should we notify the user that scope `openid` is a default scope and cannot be added or , should  we just allow client creation ignoring the `openid` scope?.
- Also, since `openid` default implicit scope, should we consider showing in the allowed list of scopes?

v2/contribute/decisions/oauth/0005-reserved-scopes.mdx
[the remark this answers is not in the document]
AdithyaSSup: @rishabhpoddar .  I will update the flow digram with the above suggested change.

v2/contribute/decisions/oauth/0007-use-jwt-access-tokens.mdx
```diff
@@ -0,0 +1,45 @@
+---
```
rishabhpoddar, line 1: Some other ADRs:
- Do we need to store the access token hash in the db at all? Right now, we store it only to be able to provide an endpoint wherein the user can validate if the access token exists in the db. But this can also be done by checking against the `id` in the access token, which is already in the db (similar to the concept of `sessionHandle` for session recipe). The advantage of this is that, if we provide an endpoint to change the access token claims (is that even a thing?) then older access tokens will continue to work even from a db checking point of view - since the `id` will not change.
- Related to the above point, should we provide an endpoint to update the access token claims? What if the user's roles change, then devs may want new access and id tokens for that user right? How do other auth providers handle this?
- We will need to add the access token and id token payload in the db cause we allow users to give custom claims in them, and post refresh, we need to be able to re add those claims in the new tokens. Any better approach other than storing the claims in the db?

v2/contribute/decisions/oauth/0004-scope-list.mdx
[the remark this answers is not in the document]
rishabhpoddar: > I don't think we even need to add it to the allowed scopes for client table

I think it's easier from a logic / db point of view to just add it always in the allowed scopes for each client.

> No, we don't really need to add it to the scope list if it exists; if the client is not requesting an id_token we should not generate one.

So if during the auth code grant flow, the scopes is an empty array, should we throw an error?

> if the user tries to add openid to the list of required scopes during client creation, should we notify the user that scope openid is a default scope with a warning or a note or , should we just allow client creation ignoring the openid scope?.

If the openid scope is a part of the allowed scopes array during client creation, we keep it in there. If it's not, we will add it. Also, instead of "list of required scopes", did you mean "list of allowed scopes"?

> Also, since openidis an implicit default scope, should we consider showing in the allowed list of scopes?
What do you mean?

v2/contribute/decisions/oauth/0005-reserved-scopes.mdx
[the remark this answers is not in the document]
rishabhpoddar: Agreed.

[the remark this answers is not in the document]
rishabhpoddar: We should add this to some ADR @porcellus. Keeping this comment open until it's in an ADR

v2/contribute/decisions/oauth/0010-m2m-admin-access.mdx
rishabhpoddar: Right yea. This conversation diverged into another topic, but we should add a profile endpoint in the backend SDK then which does OAuth 2.0 access token validation right? Should this endpoint also be able to be queried by the session recipe's access token as well?

v2/contribute/decisions/oauth/0014-not-expiring-tokens.mdx
rishabhpoddar: Oh. I'm ok with the lifetimes being fixed after creation (i.e, during refresh they can't be set to a new lifetime). What i was saying is that since users can set custom lifetimes during creation of the tokens, how will we know what those values are during the next refresh if we don't store this value in the db?

v2/contribute/decisions/oauth/0003-consent-screen.mdx
rishabhpoddar: There is already a boolean like `requires_consent` in the `oauth2_client_allowed_scopes` table which will be false for all scopes right now. Later on, when we add the consent screen feature, we can make this boolean modifiable by the dev.

rishabhpoddar (approved):

v2/contribute/decisions/oauth/0007-use-jwt-access-tokens.mdx
rishabhpoddar: Also, which code challenge methods should we support - for example, just SHA256 or more?

v2/contribute/decisions/oauth/0004-scope-list.mdx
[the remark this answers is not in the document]
porcellus: > So if during the auth code grant flow, the scopes is an empty array, should we throw an error

No, but we should only add it in the empty array/no array case.

v2/contribute/decisions/oauth/0014-not-expiring-tokens.mdx
porcellus: What I'm saying is that we don't really need to know - the user can just pass it the next time around.
Manual refreshing the token by directly calling the core is not really an intended use-case.
The refresh endpoint/functionality exposed by the backend SDK will:
1. Build an access and id token payloads based on the user/session id and the scopes (and context)
2. Pass the above payloads to the user provided callbacks if present
3. Get the token lifetimes from a user provided callback if present (otherwise falling back to the value set in the core config)
4. call the core, which will receive both the custom payloads and lifetimes (so there is no need to store them)

Keep in mind, that this is not really intended to be used the same way as our own access tokens: if you are changing the state of the user in the auth server we have no way to update these tokens for multiple reasons, so I think it's better to regenerate the tokens during refresh calls and (matching this pattern) re-generating the lifetimes as well.

Let's discuss this on a call tomorrow, then I'll add it to an ADR.

v2/contribute/decisions/oauth/0007-use-jwt-access-tokens.mdx
porcellus: > The advantage of this is that, if we provide an endpoint to change the access token claims (is that even a thing?) then older access tokens will continue to work even from a db checking point of view - since the id will not change.

The token claims could change over time, basically if you call refresh your information in the token should be updated. 
Changing the payload for the current access token has it's issues:
- we should only issue a new access token during refresh calls. 
- Since the ability to change the "current" token is missing I don't think we should add an endpoint to do it "offline".
- You can not add new scopes during refresh, which kind of means that you should not add any new information

Common use-cases, like requiring MFA can be done by redirecting the user to the auth server with additional required scopes. Keep in mind, that this may not require the user to re-enter their password, they can be just redirected to the MFA screen.

Older access tokens working isn't an advantage if you can't change them outside of the refresh endpoint. It'd basically mean that old tokens will keep working even if checking the DB directly after you called refresh.

> Any better approach other than storing the claims in the db?

I've outlined this elsewhere but let's discuss this on a call.

v2/contribute/decisions/oauth/0010-m2m-admin-access.mdx
porcellus: > we should add a profile endpoint in the backend SDK then which does OAuth 2.0 access token validation right

Yes, I'll add this to the BE SDK interface.

> Should this endpoint also be able to be queried by the session recipe's access token as well?

We can create a similar one but I don't think we should mix the two together as it will create confusion (e.g.: how do scopes relate to session recipe tokens?)

v2/contribute/decisions/oauth/0004-scope-list.mdx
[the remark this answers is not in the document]
AdithyaSSup: > Also, instead of "list of required scopes", did you mean "list of allowed scopes"?

No, I meant, if the user doesn't explicitly add `openid` scope during client creation, then in such a case should we display `openid` as one of the scopes in the create client api response?

v2/contribute/decisions/oauth/0014-not-expiring-tokens.mdx
porcellus: - Add a new core endpoint to fetch the token info based on the refresh (this will be used as an input to the payload/lifetime generator callbacks)
- Add a new column into the token info if we require a client secret when refreshing (set to true if the auth code/token was created with a client secret)
- The client secret should be required to consume an authcode if not using PKCE (thus setting the above field to true in this case)
- We are OK with the user not  being able to disallow offline access (== issuing refresh tokens) on a client level in the core. They can do this by overriding functions in the backend SDK.

v2/contribute/decisions/oauth/0004-scope-list.mdx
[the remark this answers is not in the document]
rishabhpoddar: > No, but we should only add it in the empty array/no array case.

I don't think so. If the scopes array is empty during the auth code grant flow, we should just keep it empty and NOT issue the id token. The access token will contain minimum info (just the user ID and start jwt claims). 

What do you think?

---------
Just to clarify (since this conversation is confusing now), ONLY if a CLIENT is created with an empty allowed scopes array, we automatically add the openid scope to their allowed scopes.

v2/contribute/decisions/oauth/0005-reserved-scopes.mdx
```diff
@@ -35,6 +35,12 @@ Some scopes we've decided to reserve:
 * `supertokens.roles`: access to role information 
 * `supertokens.permissions`: access to permission information
 
+Based on the above scopes we will:
+- Add information mainly into the id token
+- Make this information available through the access token calling a user info endpoint
+- Not add these to the access token (although we will allow that to be customizable)
+- Do not add any information outside of the user id (and sessionHandle) without an appropriate scope, since adding PII should always be opt-in.
```
rishabhpoddar, line 42: > Add information mainly into the id token

I was under the impression that we will add it to the access and the id token. For example, what if the dev, on their backend, has routes in which they check for roles. If the access token in this case doesn't have roles, then how will this work? Do other providers add info only to the id token and not access token?

v2/contribute/decisions/oauth/0010-m2m-admin-access.mdx
rishabhpoddar: Agreed.

v2/contribute/decisions/oauth/0018-payload-and-lifetime-after-refresh.mdx
```diff
@@ -0,0 +1,49 @@
...
+* They are re-used from when the tokens where first issued
+
+## Decision Outcome
+
+Chosen option: Token payloads and lifetimes are regenerated on the BE for refresh
+
+* Refreshing ensures fresh information
+* Simple to explain: the token is always built using the same callbacks
+
+As a sub-decision, we've discussed adding an endpoint to load the information needed for these callbacks vs adding them into the token (kind of like a JWT):
```
rishabhpoddar, line 32:
```suggestion
As a sub-decision, we've discussed adding an endpoint on the core to load the information needed for these callbacks vs adding them into the token (kind of like a JWT):
```

v2/contribute/decisions/oauth/0018-payload-and-lifetime-after-refresh.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+## Considered Options
+
+* Token payloads and lifetimes are regenerated on the BE for refresh
+* They are re-used from when the tokens where first issued
+
+## Decision Outcome
+
+Chosen option: Token payloads and lifetimes are regenerated on the BE for refresh
+
```
rishabhpoddar, line 28: related to this, will the id token's lifetime and custom payload also come in a similar way? Cause then, the API spec for the core needs to all allow to pass in idTokenLifetime and customIDTokenPayload as an input to token token POST API (similar to how we have accessTokenLifetime and customAccessTokenPayload)

v2/contribute/decisions/oauth/0019-requiring-client-secret.mdx
```diff
@@ -0,0 +1,58 @@
...
+import ArgumentNeut from "/src/components/decisionLogs/ArgumentNeut"
+
+# Client secret is required if not using PKCE
+
+<DecisionHeader status="proposed" lastUpdate="2023-05-11" created="2023-05-11" deciders={["rishabhpoddar", "porcellus"]} proposedBy={["porcellus"]} />
+
+## Context and Problem Statement
+
+In certain cases, the specs would require us to differentiate between client types (if a client secret has been issued or not). Because we are not supporting all flows, this could be simplified.
+This mainly comes down to requiring a client secret to be passed or not when createing an auth code or tokens.
```
rishabhpoddar, line 19:
```suggestion
This mainly comes down to requiring a client secret to be passed or not when creating an auth code or tokens.
```

v2/contribute/sidebars.js
```diff
@@ -271,7 +271,10 @@ module.exports = {
             "decisions/oauth/0014",
             "decisions/oauth/0015",
```
rishabhpoddar, line 272: how would things like changes in the user's role be propagated to the client's access / id token?

rishabhpoddar (changes requested):

v2/contribute/sidebars.js
porcellus: During refresh: when we are rebuilding the token payloads, the built-in claims/scopes will be refreshed as well. They are part of the payloads passed to the user callback.

v2/contribute/decisions/oauth/0018-payload-and-lifetime-after-refresh.mdx
porcellus: Custom id-token payload is already there, and we can add a separate lifetime for the id token lifetime as well if you prefer. The basic idea was that it'd get the same lifetime as the access token, but I guess the name is not as descriptive as it could be in this case.

v2/contribute/decisions/oauth/0004-scope-list.mdx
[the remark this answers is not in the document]
porcellus: This is what I actually meant, not sure how I missed the auth code grant flow part.

v2/contribute/decisions/oauth/0018-payload-and-lifetime-after-refresh.mdx
porcellus: We've decided that the id-token will be created separately, so it's lifetime will be controlled by the BE exclusively.

v2/contribute/decisions/oauth/0004-scope-list.mdx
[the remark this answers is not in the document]
rishabhpoddar: Okay. Please add to some ADR. Then i will close this comment

v2/contribute/decisions/oauth/0005-reserved-scopes.mdx
rishabhpoddar: This is very confusing. I was under the impression that the user would just add `supertokens.roles` and `supertokens.permissions` as scopes and the access token would have the list of roles / permissions for the user (just like the session token has). 

We should discuss this on a call.

v2/contribute/decisions/oauth/0005-reserved-scopes.mdx
```diff
@@ -40,6 +40,7 @@ Based on the above scopes we will:
 - Make this information available through the access token calling a user info endpoint
 - Not add these to the access token (although we will allow that to be customizable)
 - Do not add any information outside of the user id (and sessionHandle) without an appropriate scope, since adding PII should always be opt-in.
+- The main information content of the access token is contained in the scopes granted: these should be the main source of information when deciding if an operation is allowed or not. As an example: the user has an admin role, so their sessions can be used to delete users but this is not something they want to do through the forum client app. Keep in mind, that the client applications are usually not first-party, so permissions have to be carefully managed (e.g.: to mitigate a possible XSS in the forum app, even if we are assuming the app devs of the forum are not malicious).
```
rishabhpoddar, line 43: To discuss this

rishabhpoddar (changes requested):

v2/contribute/decisions/oauth/0005-reserved-scopes.mdx
rishabhpoddar: Please check how other providers do this and then we can add an ADR for this.

rishabhpoddar: Other providers by default don't add roles to the access token. Therefore, we should not do that too.

rishabhpoddar (changes requested):

PR #713, opened 2023-09-19
Update testing-with-postman.mdx

varunkakkar:
Corrected the Postman Example to use the **GET** `/sessionInfo` API where **POST** `/change-user-data` is being used

## Summary of change
The Documentation for Testing with Postman mentions the use of a **POST** `/change-user-data` API. The said API has been updated to **GET** `/sessionInfo` but the same does not reflect in the Documentation.

The documentation also states an incorrect response message to the **GET** `/sessionInfo` API which says `try refresh token` instead of `unauthorized`

This pull request updates the documentation to reflect the same.

### DIFF
--- a/v2/emailpassword/testing/testing-with-postman.mdx
+++ b/v2/emailpassword/testing/testing-with-postman.mdx
@@ -153,7 +153,7 @@ By default, for `GET` APIs, you don't need to provide the `anti-csrf` request he
 
 <img alt="Request to change user data in postman" width="600px" src="/img/^{rid}/change-user.png" />
 
-In case you query the `/sessionInfo` API with an expired access token, you will get a `401` response with the message `uauthorized`. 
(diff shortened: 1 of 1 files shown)


rishabhpoddar:
Hey @varunkakkar 

> The said API has been updated to GET /sessionInfo but the same does not reflect in the Documentation.

Where has this happened?

varunkakkar:
Hi @rishabhpoddar, I was running the backend server and testing the same with Postman for the emailpassword recipe. To do this I was following the steps in the documentation.


The documentation mentions **POST** `/change-user-data` API to verify the access token validity. Which is incorrect as the API which I could see in the backend server was **GET** `/sessionInfo`

rishabhpoddar:
The docs also mentions that you should add the `/change-user-data` to the backend server. So this isn't really any issue IMO.

rishabhpoddar:
Closing this due to no response, and cause the change is not needed.

PR #724, opened 2023-10-04
Add explanation for Flutter social login with Apple on Android

nkshah2:
## Summary of change

## Related issues
- Link to issue1 here
- Link to issue1 here

## Checklist

- [ ] Item1
- [ ] Item2

### DIFF
--- a/v2/thirdparty/custom-ui/thirdparty-login.mdx
+++ b/v2/thirdparty/custom-ui/thirdparty-login.mdx
@@ -323,6 +323,8 @@ Coming Soon
 
 <TabItem value="ios">
 
+<h3>Step 1) Fetching the authorisation token on the frontend</h3>
+
 For iOS you use the normal sign in with apple flow and then use the id token to login with SuperTokens
 
 ```swift
@@ -363,6 +365,8 @@ fileprivate class ViewController: UIViewController, ASAuthorizationControllerPre
 
 <TabItem value="flutter">
 
+<h3>Step 1) Fetching the authorisation token on the frontend</h3>
+
 For flutter we use the [sign_in_with_apple](https://pub.dev/packages/sign_in_with_apple) package, make sure to follow the prerequisites steps to get the package setup. After setup use the snippet below to trigger the apple sign in flow.
 
 ```dart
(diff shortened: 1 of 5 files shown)


v2/thirdparty/custom-ui/thirdparty-login.mdx
```diff
@@ -373,6 +373,13 @@ void loginWithApple() async {
     var credential = await SignInWithApple.getAppleIDCredential(scopes: [
       AppleIDAuthorizationScopes.email,
       AppleIDAuthorizationScopes.fullName,
+      // Required for Android only
+      webAuthenticationOptions: WebAuthenticationOptions(
+        clientId: "<CLIENT_ID>",
+        redirectUri: Uri.parse(
+          "<API_DOMAIN>/<API_BASE_PATH>/callback/apple",
+        ),
```
rishabhpoddar, line 381: something si wrong here.

v2/thirdparty/custom-ui/thirdparty-login.mdx
```diff
@@ -387,6 +394,64 @@ void loginWithApple() async {
...
 ```
 
+In the snippet above for Android we need to pass an additional `webAuthenticationOptions` property when signing in with Apple. This is because on Android the library uses the web login flow and we need to provide it with the client id and redirection uri. The `redirectUri` property here is the URL to which Apple will make a `POST` request after the user has logged in. The SuperTokens backend SDKs provide an API for this at `<API_DOMAIN>/<API_BASE_PATH>/callback/apple`.
+
+<h3> Additional steps for Android </h3>
+
+For android we also need to provide a way for the web login flow to redirect back to the app. By default the API provided by the backend SDKs redirect to the website domain you provide when initialising the SDK, we can override the API to have it redirect to our app instead. For example if you were using the Node.js SDK:
+
+```tsx
+import ^{recipeNameCapitalLetters} from "supertokens-node/recipe/^{codeImportRecipeName}";
```
rishabhpoddar, line 404: need for python and golang too

v2/thirdparty/custom-ui/thirdparty-login.mdx
```diff
@@ -373,6 +373,13 @@ void loginWithApple() async {
     var credential = await SignInWithApple.getAppleIDCredential(scopes: [
```
rishabhpoddar, line 373: in this flutter section, add Step 1)

v2/thirdparty/custom-ui/thirdparty-login.mdx
rishabhpoddar: This is also true for ios

rishabhpoddar (changes requested):

PR #729, opened 2023-10-12
Add next app dir docs for thirdpartyemailpassword

nkshah2:
## Summary of change

## Related issues
- Link to issue1 here
- Link to issue1 here

## Checklist

- [x] Add about page with CLI command
- [x] Custom UI dont need window handler and set router
- [x] Double check all example app links
- [x] When using declare for local imports we need to ignore typing for local imports (check other places where this is done and how its done there)
- [ ] At the end of it, create a next js app and copy paste all the setup steps to make sure it works
- [x] Fix logic of cache control header to add it only if it does not already exist (this needs to be done in docs, CLI and canary app)
- [x] Change description of getSSRSession to just describle the different results nd their meaning
- [x] withSession decsrption: This function will attach all session tokens to the final response if a valid session exists. -> Explain that it will return 401 if the session is expired, if the payload claims are invalid (dont pass validation) then it will return 403
- [x] Remove disclaimer for sessio nauth not working in SSR
- [x] Talk about client components before session components: http://localhost:3000/docs/thirdpartyemailpassword/nextjs/app-directory/protecting-route
- [x] Refactor comment in Using SessionAuthForNextJS and checking for sessions to split them in the different cinditions to make it clearer
- [x] Add snippet for TryRefreshComponent before the usage. Add a section to explain creating the helper components
- [x] Hello world -. Your user id is: http://localhost:3000/docs/thirdpartyemailpassword/nextjs/app-directory/protecting-route
- [x] redirect("/auth"); -> SuperTokens.redirectToAuth (in CLI, next canary and docs)
- [x] Typo not exist it wil ltry http://localhost:3000/docs/thirdpartyemailpassword/nextjs/app-directory/protecting-route
- [x] it checks if a session exists and if it does not exist it wil ltry to call the refresh endpoint remove. It just redirects
- [x] At the moment the SessionAuth component does not support server side rendering and will only work on the client side. -> caution and mention the next section (server components) for how to use sessions on server side
- [x] useSessionContext can be used along with the SessionAuth wrapper component. refactor to mention that it does not need SessionAuth
- [ ] http://localhost:3000/docs/thirdpartyemailpassword/nextjs/app-directory/protecting-route needs custom UI as well. Link to the session recipe docs for custom UI for client compoennts. Server components will be similar to pre built
- [x] Adding a session guard to all API routes -> Each API route
- [x] http://localhost:3000/docs/thirdpartyemailpassword/nextjs/app-directory/session-verification-session-guard -> mention that expired results in 401 and invalid claims results in 403
- [x] http://localhost:3000/docs/thirdpartyemailpassword/nextjs/app-directory/session-verification-middleware -> move content to the top. Also mention that the reson why the full session object is not added is because nextjs middlewares dont allow you to pass objects to your routes. Also add example of API using this information. Add a disclaimer that this is an alternative to session guards and that you just need to use one of thm not both (add this at the top)
- [x] Lets modify the Home page we made in a previous step to make a call to this API add a link to the API we made in the previous setp
http://localhost:3000/docs/thirdpartyemailpassword/nextjs/app-directory/server-components-requests instead of hello world use some session information
- [x] userInfoResponse check status being 200 and show user id. If 500 show something went wrong. if 401 show try refresh. If 403 show sessionauthwrapper with no children
- [ ] http://localhost:3000/docs/thirdpartyemailpassword/nextjs/app-directory/server-components-requests add custom UI. SessionAuthForNextJS needs to be manually handled and status 403 will have to be manually handled


v2/thirdpartyemailpassword/nextjs/app-directory/session-verification.mdx
```diff
@@ -0,0 +1,255 @@
...
+:::warning
+We do not recommend this approach because information set by the middleware will not change based on changes in the user session
+:::
+
+```tsx title="middleware.tsx"
+import { NextResponse } from 'next/server'
+import type { NextRequest } from 'next/server'
+import { SessionContainer } from 'supertokens-node/recipe/session'
+import { withSession } from './app/sessionUtils';
+
```
rishabhpoddar, line 219: also talk about how the API can consume the user ID then.

rishabhpoddar (changes requested):

rishabhpoddar (changes requested):

PR #732, opened 2023-10-20
Fix incorrect instructions for Next.js app router

itsmingjie:
## Summary of change

In App Router, the correct way to create a catch-all segment is through `[[...segmentName]]/page.tsx`, as you can no longer use dynamic routing on the leaf of directories. [Supertokens' app directory example](https://github.com/supertokens/next.js/blob/canary/examples/with-supertokens/app/auth/%5B%5B...path%5D%5D/page.tsx) is correct. 

## Related commits

Seems like the example was also super new, documentation typo, I presume. 

- https://github.com/supertokens/next.js/commit/9e7d2d011121dd6fa90c491c9e81ee7c7552b710#diff-401e3d26e1b3958c21cd08b9146e1e9cc9f4db8652c259644b96693e433217f5

## Checklist


nkshah2:
Hi @itsmingjie 

Thanks for the PR, Ive updated your branch to also include the changes you made in other recipes,

nkshah2 (approved):

PR #778, opened 2024-02-29
Update claim-validators.mdx

JakubZojdzik:
## Summary of change
Change markdown text blocks to blocks of tsx code when needed to fix syntax highlighting


rishabhpoddar:
Hey. We can't merge this cause the code snippet is not a correct type script snippet (i.e. it does not compile correctly with ts), and adding tsx instead of text will make our build flow try and compile it with ts, which will fail.

PR #783, opened 2024-03-20
fix: import statements and other general improvements to the nestjs g…

mawentowski:
## Summary of change
Improving the NestJS guide

## Checklist

### DIFF
--- a/v2/thirdpartyemailpassword/nestjs/guide.mdx
+++ b/v2/thirdpartyemailpassword/nestjs/guide.mdx
@@ -4,13 +4,14 @@ title: Guide
 hide_title: true
 ---
 
+<!-- COPY SECTION -->
+<!-- ./thirdpartyemailpassword/nestjs/guide.mdx -->
+<!-- 1 -->
 import AppInfoForm from "/src/components/appInfoForm"
 import CoreInjector from "/src/components/coreInjector"
-import {Question, Answer}from "/src/components/question"
+import {Question, Answer} from "/src/components/question"
 
-# Guide
-
-## Overview
+# NestJS Integration Guide
 
 Integrating SuperTokens into a NestJS backend is a bit different than the quick setup guide shows. We will add a few things:
 - A module to house all authorization-related code
@@ -24,13 +25,13 @@ We will cover each of these in the following few sections. Then, you can do the 
 
 Please look [here](https://docs.nestjs.com/first-steps) to see how to get started with your NestJS backend.
 
-## 1) Installing SuperTokens
+## Installing SuperTokens
 
 ```bash
 npm i -s supertokens-node
 ```
 
-## 2) Adding a new module
+## Adding a new module
 
 You can scaffold a module using the nest CLI by running this in the root folder of the application:
 ```bash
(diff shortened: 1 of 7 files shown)


v2/thirdpartyemailpassword/nestjs/guide.mdx
```diff
@@ -36,16 +36,16 @@ You can scaffold a module using the nest CLI by running this in the root folder
...
+The result should be a new `auth` folder with `auth.module.ts` in it. 
 
 ### Add config type and injection token
 
-Add a `config.interface.ts` file next into the `auth` folder. We will put the type and injection token for the SuperTokens config here.
+Create a `config.interface.ts` file in the `auth` folder. We will put the type and injection token for the SuperTokens configuration here.
 
-```ts
-import { AppInfo } from "supertokens-node/types";
+```ts title="./src/config.interface.ts"
```
rishabhpoddar, line 45: this is not the rigth path, since the line above says in the auth folder, but the title of this snippet says src/config..

v2/thirdpartyemailpassword/nestjs/guide.mdx
```diff
@@ -36,16 +36,16 @@ You can scaffold a module using the nest CLI by running this in the root folder
 ```bash
 nest g module auth
 ```
-The result should be a new `auth` folder with `auth.module.ts` in it. We should convert this into a [dynamic module](https://docs.nestjs.com/modules#dynamic-modules) so we can set parts of the SuperTokens configuration in the App module. Centralizing settings like this can be helpful for things like using a separate connection URI for testing.
+The result should be a new `auth` folder with `auth.module.ts` in it. 
 
 ### Add config type and injection token
 
-Add a `config.interface.ts` file next into the `auth` folder. We will put the type and injection token for the SuperTokens config here.
+Create a `config.interface.ts` file in the `auth` folder. We will put the type and injection token for the SuperTokens configuration here.
```
rishabhpoddar, line 44: since the time this guide was written, nestjs did not create sub folders for `auth/auth` for example. All the commands like `nest g service supertokens auth` or `nest g middleware auth auth` would create the files inside the `./src/auth` folder instead. So we would not have files like `./src/auth/auth/auth.filter.spec.ts`, instead, it was just `./src/auth/auth.filter.spec.ts`. 

Can you figure out how we can make it so that this extra folder is not created? It's really strange to have ./src/auth/auth/...

rishabhpoddar (changes requested):

v2/thirdpartyemailpassword/nestjs/guide.mdx
mawentowski: I changed this file to be located in `./src/auth`.

v2/thirdpartyemailpassword/nestjs/guide.mdx
mawentowski: The behavior I see, for example: `nest g middleware auth,` is that it knows it knows to put that file in the `auth` directory from the root (i.e., it does not need an extra `auth` in the command). When you add the extra `auth`, it nests it in `auth/auth`. So, I removed the double auth occurrences in the commands.

So the `auth` files are now located as such:

```txt
📦src
 ┣ 📂auth
 ┃ ┣ 📂supertokens
 ┃ ┃ ┣ 📜supertokens.service.spec.ts
 ┃ ┃ ┗ 📜supertokens.service.ts
 ┃ ┣ 📜auth.filter.spec.ts
 ┃ ┣ 📜auth.filter.ts
 ┃ ┣ 📜auth.guard.spec.ts
 ┃ ┣ 📜auth.guard.ts
 ┃ ┣ 📜auth.middleware.spec.ts
 ┃ ┣ 📜auth.middleware.ts
 ┃ ┣ 📜auth.module.ts
 ┃ ┗ 📜config.interface.ts
 ┣ 📜app.controller.spec.ts
 ┣ 📜app.controller.ts
 ┣ 📜app.module.ts
 ┣ 📜app.service.ts
 ┗ 📜main.ts
```

PR #790, opened 2024-04-17
feat: Add bulk import docs

anku255:
## Summary of change

## Related issues

- https://github.com/supertokens/supertokens-core/issues/912

## Checklist

### DIFF
--- a/v2/emailpassword/migration/about.mdx
+++ b/v2/emailpassword/migration/about.mdx
@@ -9,42 +9,27 @@ hide_title: true
 
 # About
 
-In this guide we will be going through the process of migrating users from an external Authentication provider to SuperTokens.
+This guide explains how to migrate users from an external authentication provider to SuperTokens.
 
-There are 3 steps to user migration:
+Migration involves two main steps:
 
-- Account Migration
-  - User Creation
-  - UserId Mapping
-  - Mark email as verified
-- User Data Migration
-- Session Migration
+Account Migration
+Session Migration
 
 ## Step 1) Account Migration:
 
-### User Creation
-- Our first step involves creating a SuperTokens user by importing their account credentials from the previous authentication provider.
-- You can learn more about how to implement these changes in the [User Creation](./account-creation/user-creation) section. 
+In the first step, use the Bulk User Import API to create a SuperTokens user for each user in your existing authentication provider. This API allows you to:
 
-### User ID Mapping
-- If you have stored information against existing userIds in your application table, you can use UserId Mapping to map the existing userId to the user's SuperTokens userId.
-- Once the userIds are mapped you can use the existing userId to interact with all of SuperTokens APIs.
-- You can learn more about how to implement these changes in the [User Id Mapping](./account-creation/user-id-mapping) section.
+- Create a SuperTokens user.
+- Map the existing userId to the SuperTokens userId.
+- Mark the email as verified (if applicable).
+- Add roles, metadata, and MFA devices (if applicable).
+- Perform account linking (if applicable).
 
-### Mark email as verified
-- Once a SuperTokens user has been created and their userId has been mapped, you need to mark their email as verified (if applicable) so that  they do not have to go through the email verification process again.
-- You can learn more about how to implement these changes in [this section](./account-creation/email-verification).
+For implementation details, refer to the [Bulk User Import](./bulk-user-import) section.
 
-## Step 2) User Data Migration
-- Now that your user's account has been migrated over to SuperTokens we can associate additional information like roles and metadata with your user.
--  You can learn more about how to implement these changes in the [User Data Migration](./data-migration) section.
+## Step 2) Session Migration
 
-## Step 3) Session Migration
+If users have active sessions, use [this guide](./session-migration) to migrate their external provider sessions to SuperTokens sessions. This way, users won't need to re-authenticate.
 
-- If you have users with an existing session, you can use [this guide](./session-migration) to migrate their external provider sessions to a SuperTokens session.
-- This will prevent users from having to re-authenticate.
-
-You can learn more about how to implement these changes in the [Session Migration](./session-migration) section.
-
-## Step 4) MFA migration
-If you are using MFA in your app, checkout the MFA migration section [here](/docs/mfa/migration) after you have gone through the above migration steps.
+Learn how to implement session migration in the [Session Migration](./session-migration) section.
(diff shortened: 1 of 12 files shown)


v2/emailpassword/migration/about.mdx
```diff
@@ -9,42 +9,27 @@ hide_title: true
...
+Migration involves two main steps:
 
-- Account Migration
-  - User Creation
-  - UserId Mapping
-  - Mark email as verified
-- User Data Migration
-- Session Migration
+Account Migration
+Session Migration
```
sattvikc, line 17: not properly formatted

v2/emailpassword/migration/about.mdx
```diff
@@ -9,42 +9,27 @@ hide_title: true
...
 
-### User ID Mapping
-- If you have stored information against existing userIds in your application table, you can use UserId Mapping to map the existing userId to the user's SuperTokens userId.
-- Once the userIds are mapped you can use the existing userId to interact with all of SuperTokens APIs.
-- You can learn more about how to implement these changes in the [User Id Mapping](./account-creation/user-id-mapping) section.
+- Create a SuperTokens user.
+- Map the existing userId to the SuperTokens userId.
+- Mark the email as verified (if applicable).
+- Add roles, metadata, and MFA devices (if applicable).
+- Perform account linking (if applicable).
```
sattvikc, line 27: You should probably say linking multiple login methods for same user. Not sure if people understand what account linking would mean.

v2/emailpassword/migration/about.mdx
```diff
@@ -9,42 +9,27 @@ hide_title: true
 
 # About
 
-In this guide we will be going through the process of migrating users from an external Authentication provider to SuperTokens.
+This guide explains how to migrate users from an external authentication provider to SuperTokens.
```
sattvikc, line 12: migrate users from a different authentication provider such as `Auth0`, ... etc to SuperTokens.

v2/thirdpartyemailpassword/migration/bulk-user-import.mdx
```diff
@@ -0,0 +1,253 @@
...
+Convert your existing authentication provider data to the required format as specified [here](./bulk-user-import-schema-examples.mdx). An example JSON would look like this:
+
+```json
+{
+  "users": [
+    {
+      "loginMethods": [
+        {
+          "recipeId": "emailpassword",
+          "tenantIds": ["public"],
```
sattvikc, line 33: should `tenantIds` with `["public"]` be a mandatory input? this would be applicable only for importing multi-tenant users right?

v2/thirdpartyemailpassword/migration/bulk-user-import.mdx
```diff
@@ -0,0 +1,253 @@
...
+          "isPrimary": false,
+          "email": "johndoe@gmail.com",
+          "thirdPartyId": "google",
+          "thirdPartyUserId": "106347997792363870000",
+          "timeJoinedInMSSinceEpoch": 1713260578868
+        }
+      ]
+    }
+  ]
+}'
```
sattvikc, line 219: have you tested this command? seems like the password hash contain $ might try to resolve to an environment variable. Please ensure this works fine.

v2/thirdpartyemailpassword/migration/bulk-user-import-schema-examples.mdx
```diff
@@ -0,0 +1,200 @@
...
+# Bulk User Import Schema and Examples
+
+
+The bulk import API expects the payload in the following format. 
+
+```json
+{
+  "users": {
+    "type": "array",
+    "items": {
```
sattvikc, line 16: Is it possible to render this in some way? it's way too confusing. People who don't know about json schema, won't even understand this.

v2/thirdpartyemailpassword/migration/bulk-user-import-schema-examples.mdx
```diff
@@ -0,0 +1,200 @@
...
+      },
+      "required": ["loginMethods"]
+    }
+  }
+
+```
+
+## Examples
+
+### Example 1: Sample Bulk User Import JSON
```
sattvikc, line 145: I think you must provide many more examples. All in one examples might not be very effective. It should be a step by step additions.

For example:
Start with a very simple use case, of importing emailpassword users only. then, thirdparty users.
Then probably add an example on how to assign user roles. Later metadata.
Then give examples of account linking where same user has different login methods. Talk about isPrimary. Talk about where to add user roles and metadata.

Think about this in more detail please.

sattvikc (changes requested):

v2/thirdpartyemailpassword/migration/bulk-user-import.mdx
```diff
@@ -0,0 +1,253 @@
+---
+id: bulk-user-import
+title: Bulk User Import
```
sattvikc, line 3: This page should probably called user accounts migration

v2/thirdpartyemailpassword/migration/bulk-user-import.mdx
```diff
@@ -0,0 +1,253 @@
...
+      ]
+    }
+  ]
+}'
+```
+</CoreInjector>
+
+## Step 3: Monitor the import status
+
+Monitor the import status with a GET request to the bulk import API. Filter the response using the `status` query parameter, which can be 'NEW', 'PROCESSING', or 'FAILED'. Completed entries are automatically deleted.
```
sattvikc, line 225: You should talk about when the import is going to happen. This is because it won't happen as soon as the bulk-import API is called. It should be clear on when the user can check for the statuses. Also is it possible for the user to know if the users list he passed has been processed or not?

sattvikc (changes requested):

v2/emailpassword/migration/about.mdx
anku255: We have used the term "account linking" in our docs before too.

v2/thirdpartyemailpassword/migration/bulk-user-import.mdx
anku255: No, it is not a mandatory input. I only included it make it obvious that the users will be imported to the `public` tenant.

v2/thirdpartyemailpassword/migration/bulk-user-import.mdx
anku255: Yeah, it works. 

image

PR #791, opened 2024-04-23
fix: Update docs for olderCookieDomain config

anku255:
## Summary of change
This PR updates the docs for `olderCookieDomain` config option in backend SDKs.

## Related issues

- https://github.com/supertokens/supertokens-node/pull/813

## Checklist

### DIFF
--- a/v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
+++ b/v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
@@ -52,6 +52,7 @@ SuperTokens.init({
         Session.init({
             // highlight-next-line
             cookieDomain: ".example.com",
+            olderCookieDomain: "" // Set to an empty string if your previous cookieDomain was unset. Otherwise, use your old cookieDomain value.
         })
     ]
 });
@@ -69,12 +70,14 @@ import (
 func main() {
 	// highlight-next-line
 	cookieDomain := ".example.com"
+	olderCookieDomain := "" // Set to an empty string if your previous cookieDomain was unset. Otherwise, use your old cookieDomain value.
 
 	supertokens.Init(supertokens.TypeInput{
 		RecipeList: []supertokens.Recipe{
 			session.Init(&sessmodels.TypeInput{
 				// highlight-next-line
 				CookieDomain: &cookieDomain,
+				OlderCookieDomain: &olderCookieDomain,
 			}),
 		},
 	})
@@ -95,6 +98,7 @@ init(
         session.init(
             # highlight-start
             cookie_domain='.example.com'
+            older_cookie_domain='' # Set to an empty string if your previous cookie_domain was unset. Otherwise, use your old cookie_domain value.
             # highlight-end
         )
     ]
@@ -110,6 +114,10 @@ The above will set the session cookies' domain to `.example.com`, allowing them 
 Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDomain` in `appInfo` must point to an exact API domain only. This should be the API in which you want to expose all the auth related endpoints (for example `/auth/signin`).
 
 For local development, you should not set the `cookieDomain` to an IP address based domain, or `.localhost` - browsers will reject these cookies. Instead, you should [alias `localhost` to a named domain and use that](https://superuser.com/questions/152146/how-to-alias-a-hostname-on-mac-osx). 
+:::
+
+:::important
+The `olderCookieDomain` value should be set to prevent clients from having multiple session cookies from different domains. This can happen when cookies from a previous domain are still valid and sent with requests. For instance, if your previous `cookieDomain` was `api.example.com` and the new one is `.example.com`, both sets of cookies would be sent to the apiDomain `api.example.com`, leading to an inconsistent state. This can cause issues until the older cookies are cleared. Setting `olderCookieDomain` in the configuration ensures that the SuperTokens SDK can automatically remove these older cookies.
 :::
 
 ## Step 2) Frontend config
--- a/v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
+++ b/v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
@@ -28,9 +28,9 @@ To enable use of sessions for multiple API endpoints, you need to use the `sessi
 - `sessionTokenBackendDomain` in the frontend config must match the `cookieDomain` set in the backend config.
 :::
 
-## Step 1) Backend config
+## Step 1) Set Cookie Domain in the Backend Config
 
-You need to set the `sessionTokenBackendDomain` value to be the common top level domain. For example, if your API endpoints are `{"api.example.com", "api2.example.com", "api3.example.com"}`, the common portion of these endpoints is `".example.com"` (The dot is important). So you would need to set the following:
+You need to set the `cookieDomain` value to be the common top level domain. For example, if your API endpoints are `{"api.example.com", "api2.example.com", "api3.example.com"}`, the common portion of these endpoints is `".example.com"` (The dot is important). So you would need to set the following:
 
 <BackendSDKTabs>
 <TabItem value="nodejs">
@@ -52,6 +52,95 @@ SuperTokens.init({
         Session.init({
             // highlight-next-line
             cookieDomain: ".example.com",
+        })
+    ]
+});
+```
+</TabItem>
+<TabItem value="go">
+
+```go
+import (
+	"github.com/supertokens/supertokens-golang/recipe/session"
+	"github.com/supertokens/supertokens-golang/recipe/session/sessmodels"
+	"github.com/supertokens/supertokens-golang/supertokens"
+)
+
+func main() {
+	// highlight-next-line
+	cookieDomain := ".example.com"
+
+	supertokens.Init(supertokens.TypeInput{
+		RecipeList: []supertokens.Recipe{
+			session.Init(&sessmodels.TypeInput{
+				// highlight-next-line
+				CookieDomain: &cookieDomain,
+			}),
+		},
+	})
+}
+```
+</TabItem>
+<TabItem value="python">
+
+```python
+from supertokens_python import init, InputAppInfo
+from supertokens_python.recipe import session
+
+
+init(
+    app_info=InputAppInfo(api_domain="...", app_name="...", website_domain="..."),
+    framework='...', # type: ignore
+    recipe_list=[
+        session.init(
+            # highlight-start
+            cookie_domain='.example.com'
+            # highlight-end
+        )
+    ]
+)
+```
+
+</TabItem>
+</BackendSDKTabs>
+
+The above will set the session cookies' domain to `.example.com`, allowing them to be sent to `*.example.com`.
+
+:::note
+Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDomain` in `appInfo` must point to an exact API domain only. This should be the API in which you want to expose all the auth related endpoints (for example `/auth/signin`).
+
+For local development, you should not set the `cookieDomain` to an IP address based domain, or `.localhost` - browsers will reject these cookies. Instead, you should [alias `localhost` to a named domain and use that](https://superuser.com/questions/152146/how-to-alias-a-hostname-on-mac-osx). 
+:::
+
+## Step 2) Set Older Cookie Domain in the Backend Config
+
+:::caution
+If `olderCookieDomain` isn't set, users with older sessions will be locked out until the `olderCookieDomain` is set correctly or they clear their cookies.
+:::
+
+To avoid locking out users with existing sessions, set `olderCookieDomain` to match your previous `cookieDomain`. If your `cookieDomain` was not set, you can use an empty string. However, if you don't have any existing sessions, you can skip this step entirely.
+
+<BackendSDKTabs>
+<TabItem value="nodejs">
+
+```tsx
+import SuperTokens from "supertokens-node";
+import Session from "supertokens-node/recipe/session";
+
+SuperTokens.init({
+    supertokens: {
+        connectionURI: "...",
+    },
+    appInfo: {
+        apiDomain: "...",
+        appName: "...",
+        websiteDomain: "..."
+    },
+    recipeList: [
+        Session.init({
+            // highlight-next-line
+            cookieDomain: ".example.com",
+            // highlight-next-line
             olderCookieDomain: "" // Set to an empty string if your previous cookieDomain was unset. Otherwise, use your old cookieDomain value.
         })
     ]
@@ -70,14 +159,15 @@ import (
 func main() {
 	// highlight-next-line
 	cookieDomain := ".example.com"
-	olderCookieDomain := "" // Set to an empty string if your previous cookieDomain was unset. Otherwise, use your old cookieDomain value.
+    olderCookieDomain := "" // Set to an empty string if your previous cookieDomain was unset. Otherwise, use your old cookieDomain value.
 
 	supertokens.Init(supertokens.TypeInput{
 		RecipeList: []supertokens.Recipe{
 			session.Init(&sessmodels.TypeInput{
 				// highlight-next-line
 				CookieDomain: &cookieDomain,
-				OlderCookieDomain: &olderCookieDomain,
+                // highlight-next-line
+                OlderCookieDomain: &olderCookieDomain
 			}),
 		},
 	})
@@ -108,19 +198,15 @@ init(
 </TabItem>
 </BackendSDKTabs>
 
-The above will set the session cookies' domain to `.example.com`, allowing them to be sent to `*.example.com`.
-
-:::note
-Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDomain` in `appInfo` must point to an exact API domain only. This should be the API in which you want to expose all the auth related endpoints (for example `/auth/signin`).
-
-For local development, you should not set the `cookieDomain` to an IP address based domain, or `.localhost` - browsers will reject these cookies. Instead, you should [alias `localhost` to a named domain and use that](https://superuser.com/questions/152146/how-to-alias-a-hostname-on-mac-osx). 
-:::
-
-:::important
+:::info
 The `olderCookieDomain` value should be set to prevent clients from having multiple session cookies from different domains. This can happen when cookies from a previous domain are still valid and sent with requests. For instance, if your previous `cookieDomain` was `api.example.com` and the new one is `.example.com`, both sets of cookies would be sent to the apiDomain `api.example.com`, leading to an inconsistent state. This can cause issues until the older cookies are cleared. Setting `olderCookieDomain` in the configuration ensures that the SuperTokens SDK can automatically remove these older cookies.
 :::
 
-## Step 2) Frontend config
+:::info
+Changing the `cookieDomain` can cause a temporary spike in requests, even if the `olderCookieDomain` is set correctly. This happens because older sessions, with older cookie domain, require additional refresh calls to clear their old cookies and set new ones. This spike is a one-time event and should not recur after the update.
+:::
+
+## Step 3) Frontend config
 
 You need to set the same value for `sessionTokenBackendDomain` on the frontend. This will allow the frontend SDK to apply interception and automatic refreshing across all your API calls:
 
--- a/v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
+++ b/v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
@@ -114,14 +114,6 @@ For local development, you should not set the `cookieDomain` to an IP address ba
 
 ## Step 2) Set Older Cookie Domain in the Backend Config
 
-:::caution
-- If `olderCookieDomain` isn't set, users with older sessions will get a 500 error from the session refresh endpoint, effectively locking them out. This will continue until `olderCookieDomain` is set correctly or they clear their cookies.
-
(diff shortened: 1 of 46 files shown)


v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -112,6 +116,10 @@ Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDom
 For local development, you should not set the `cookieDomain` to an IP address based domain, or `.localhost` - browsers will reject these cookies. Instead, you should [alias `localhost` to a named domain and use that](https://superuser.com/questions/152146/how-to-alias-a-hostname-on-mac-osx). 
 :::
 
+:::important
```
porcellus, line 119: I think this should be moved above the snippet or merged into the important section on the top of the page.
Also, it feels like it's missing information:
- changing the cookieDomain can cause elevated request numbers even if olderCookieDomain is set right
- if olderCookieDomain is not set, it'll lock people with older sessions out
- if olderCookieDomain is set wrong it causes an infinite refresh loop (we might not want to mention this though)
- you only need to set it if you have previous sessions.

Some of these are kind of like warnings, so consider separating those into a danger/warning admonition (if we use those elsewhere). What we have here is kind of an explanation section.

v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -52,6 +52,7 @@ SuperTokens.init({
         Session.init({
             // highlight-next-line
             cookieDomain: ".example.com",
+            olderCookieDomain: "" // Set to an empty string if your previous cookieDomain was unset. Otherwise, use your old cookieDomain value.
```
porcellus, line 55: Defaulting to undefined (or commented out or whatever) could be better. Also consider something like this (not the right wording and I'm not sure if it's better anyway):
```
// setting olderCookieDomain can be skipped if you will not have older session
// olderCookieDomain: "", // Set this to an empty string if the your previous cookieDomain was unset
// olderCookieDomain: "api.example.com" // Set this to the previous cookieDomain value if it was set
```

porcellus: Another option is to add setting the `olderCookieDomain` into a separate step as that gives you a bit more room for explanation and to explain all the implications

v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
porcellus: If you choose to add a separate step for setting `olderCookieDomain`, then the warnings can go to the top of that, with the explanation (what/how it causes issues) moved into normal text or a `note` admonition.

porcellus (changes requested):

v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
anku255: I have updated the PR to add a separate step for setting `olderCookieDomain` and added some warning/info as well. Please take a look

v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -112,7 +112,101 @@ Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDom
...
+```
+
+</TabItem>
+</BackendSDKTabs>
+
+:::info
+The `olderCookieDomain` value should be set to prevent clients from having multiple session cookies from different domains. This can happen when cookies from a previous domain are still valid and sent with requests. For instance, if your previous `cookieDomain` was `api.example.com` and the new one is `.example.com`, both sets of cookies would be sent to the apiDomain `api.example.com`, leading to an inconsistent state. This can cause issues until the older cookies are cleared. Setting `olderCookieDomain` in the configuration ensures that the SuperTokens SDK can automatically remove these older cookies.
+:::
+
+:::info
```
porcellus, line 205: Having 2 info admonitions next to each other feels a bit strange. We could move the one below to the end of step 1, but I'm not sure if that's any better, so passing this on to @rishabhpoddar

v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -28,9 +28,9 @@ To enable use of sessions for multiple API endpoints, you need to use the `sessi
 - `sessionTokenBackendDomain` in the frontend config must match the `cookieDomain` set in the backend config.
```
rishabhpoddar, line 28: not related to this file, but, i think there is a section in the docs that talks about session error handlers and how to configure them. If such a section exists,. we need to modify it to also talk about the new type of error for duplicate cookie

v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -28,9 +28,9 @@ To enable use of sessions for multiple API endpoints, you need to use the `sessi
...
 :::
```
rishabhpoddar, line 29: not related to this page, but the docs for sharing sessions across sub domains also needs changing to say that even if you don't add a `.`, it will still share cookies across all frontend sub domains and their sub domains.

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -28,9 +28,9 @@ To enable use of sessions for multiple API endpoints, you need to use the `sessi
 - `sessionTokenBackendDomain` in the frontend config must match the `cookieDomain` set in the backend config.
 :::
 
-## Step 1) Backend config
+## Step 1) Set Cookie Domain in the Backend Config
 
-You need to set the `sessionTokenBackendDomain` value to be the common top level domain. For example, if your API endpoints are `{"api.example.com", "api2.example.com", "api3.example.com"}`, the common portion of these endpoints is `".example.com"` (The dot is important). So you would need to set the following:
+You need to set the `cookieDomain` value to be the common top level domain. For example, if your API endpoints are `{"api.example.com", "api2.example.com", "api3.example.com"}`, the common portion of these endpoints is `".example.com"` (The dot is important). So you would need to set the following:
```
rishabhpoddar, line 33:
```suggestion
You need to set the `cookieDomain` value to be the common top level domain. For example, if your API endpoints are `{"api.example.com", "api2.example.com", "api3.example.com"}`, the common portion of these endpoints is `".example.com"` (The dot is NOT important). So you would need to set the following:
```

rishabhpoddar: The code snippets below must not have a `.` in the value set for cookie domain

rishabhpoddar: And also the text `The above will set the session cookies' domain to .example.com, allowing them to be sent to *.example.com.` will have to change (this is below the code snippet)

rishabhpoddar: The note that is below in this section will also need to change since it says "whilst the cookieDomain can start with a leading `.` ...), but in the snippet we won't be writing it with a leading dot.

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -112,7 +112,101 @@ Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDom
 For local development, you should not set the `cookieDomain` to an IP address based domain, or `.localhost` - browsers will reject these cookies. Instead, you should [alias `localhost` to a named domain and use that](https://superuser.com/questions/152146/how-to-alias-a-hostname-on-mac-osx). 
 :::
 
-## Step 2) Frontend config
+## Step 2) Set Older Cookie Domain in the Backend Config
+
+:::caution
+If `olderCookieDomain` isn't set, users with older sessions will be locked out until the `olderCookieDomain` is set correctly or they clear their cookies.
```
rishabhpoddar, line 118: define what locked out means -> the session refresh endpoint will throw a 500 error with a message.

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -112,7 +112,101 @@ Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDom
...
+    },
+    appInfo: {
+        apiDomain: "...",
+        appName: "...",
+        websiteDomain: "..."
+    },
+    recipeList: [
+        Session.init({
+            // highlight-next-line
+            cookieDomain: ".example.com",
```
rishabhpoddar, line 142:
```suggestion
            cookieDomain: "example.com", // or ".example.com" (it's the same thing)
```

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -112,7 +112,101 @@ Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDom
...
+```go
+import (
+	"github.com/supertokens/supertokens-golang/recipe/session"
+	"github.com/supertokens/supertokens-golang/recipe/session/sessmodels"
+	"github.com/supertokens/supertokens-golang/supertokens"
+)
+
+func main() {
+	// highlight-next-line
+	cookieDomain := ".example.com"
```
rishabhpoddar, line 161:
```suggestion
	cookieDomain := "example.com"  // or ".example.com" (it's the same thing)
```

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -112,7 +112,101 @@ Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDom
...
+    olderCookieDomain := "" // Set to an empty string if your previous cookieDomain was unset. Otherwise, use your old cookieDomain value.
```
rishabhpoddar, line 162: formatting issue?

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -112,7 +112,101 @@ Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDom
...
+from supertokens_python.recipe import session
+
+
+init(
+    app_info=InputAppInfo(api_domain="...", app_name="...", website_domain="..."),
+    framework='...', # type: ignore
+    recipe_list=[
+        session.init(
+            # highlight-start
+            cookie_domain='.example.com'
```
rishabhpoddar, line 190:
```suggestion
            cookie_domain='example.com', # or ".example.com" (it's the same thing)
```

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -112,7 +112,101 @@ Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDom
...
+
+</TabItem>
+</BackendSDKTabs>
+
+:::info
+The `olderCookieDomain` value should be set to prevent clients from having multiple session cookies from different domains. This can happen when cookies from a previous domain are still valid and sent with requests. For instance, if your previous `cookieDomain` was `api.example.com` and the new one is `.example.com`, both sets of cookies would be sent to the apiDomain `api.example.com`, leading to an inconsistent state. This can cause issues until the older cookies are cleared. Setting `olderCookieDomain` in the configuration ensures that the SuperTokens SDK can automatically remove these older cookies.
+:::
+
+:::info
+Changing the `cookieDomain` can cause a temporary spike in requests, even if the `olderCookieDomain` is set correctly. This happens because older sessions, with older cookie domain, require additional refresh calls to clear their old cookies and set new ones. This spike is a one-time event and should not recur after the update.
```
rishabhpoddar, line 206: same spike in the refresh API call

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -112,7 +112,101 @@ Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDom
...
+:::
+
```
rishabhpoddar, line 208: Instead of having two info boxes one below the other, have one info box, with bullet points.

v2/emailpassword/nextjs/app-directory/protecting-route.mdx
```diff
@@ -240,7 +240,7 @@ export async function HomePage() {
              * This means that the session does not exist but we have session tokens for the user. In this case
              * the `TryRefreshComponent` will try to refresh the session.
              */
-            return <TryRefreshComponent />;
+            return <TryRefreshComponent key={Date.now()} />;
```
rishabhpoddar, line 243: add comment explaining why this is needed

v2/emailpassword/nextjs/app-directory/server-components-requests.mdx
```diff
@@ -66,7 +66,7 @@ export async function HomePage() {
         if (hasInvalidClaims) {
             return <SessionAuthForNextJS />;
         } else {
-            return <TryRefreshComponent />;
+            return <TryRefreshComponent key={Date.now()} />;
```
rishabhpoddar, line 69: add comment explaining why this is needed everywhere

rishabhpoddar (changes requested):

v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
rishabhpoddar: We can make it one admonition with bullet points:
```
:::info
- point 1
- point 2
:::
```

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -112,7 +112,101 @@ Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDom
 For local development, you should not set the `cookieDomain` to an IP address based domain, or `.localhost` - browsers will reject these cookies. Instead, you should [alias `localhost` to a named domain and use that](https://superuser.com/questions/152146/how-to-alias-a-hostname-on-mac-osx). 
 :::
 
-## Step 2) Frontend config
+## Step 2) Set Older Cookie Domain in the Backend Config
```
rishabhpoddar, line 115: we should have some caution thing that says:
- The value set for olderCookieDomain must be kept for 1 year because the cookie lifetime of the access token on the frontend is 1 year.
- If you have changed the cookieDomain more than once within one year, to prevent a stuck state, switch to header based auth for all your clients (https://supertokens.com/docs/session/common-customizations/sessions/token-transfer-method#backend-configuration-optional). The important thing here is that you have to set the backend config to header even though that doc says it's optional. This is so that all clients are forced to use header based auth.

rishabhpoddar: This brings about one point of discussion. Let's say that an app has users in this situation where they already have cookies from domain `b.example.com` and `example.com`, and now the new cookieDomain is `a.b.example.com`. In this case, there would be three tokens coming to their APIs, and we would just clear one of the older ones instead of both the older ones. Would switching to header based auth (via backend header config) as mentioned in the above actually immediately log them out and allow them to login via header based auth (therefore solving this issue)?

rishabhpoddar (changes requested):

v2/emailpassword/common-customizations/sessions/multiple-api-endpoints.mdx
anku255: done

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
anku255: done

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
anku255: done

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
anku255: done

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
anku255: done

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
anku255: What did you mean? The request spike will not be limited to refresh api call only. It will be like this - `getSession` -> `refreshSession` -> `getSession` -> `refreshSession` -> `getSession`

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
anku255: done

v2/emailpassword/nextjs/app-directory/protecting-route.mdx
anku255: done

v2/emailpassword/nextjs/app-directory/server-components-requests.mdx
anku255: done

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
anku255: 1. Done.

2. I am testing this. The logout -> signin -> getSession part is working fine after switched to header based auth. However, in the refresh api call, the accessToken is sent  instead of the refreshToken causing a logout. I am looking into this. I don't think this is related to cookie domain change. Could be some bad config on my end.

rishabhpoddar: when you say accessToken is sent instead of refresh token, you mean as an authorization bearer token? How did that happen? I mean even if it’s a config error, what config error caused this?

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -104,15 +104,111 @@ init(
...
 Whilst the `cookieDomain` can start with a leading `.`, the value of the `apiDomain` in `appInfo` must point to an exact API domain only. This should be the API in which you want to expose all the auth related endpoints (for example `/auth/signin`).
 
 For local development, you should not set the `cookieDomain` to an IP address based domain, or `.localhost` - browsers will reject these cookies. Instead, you should [alias `localhost` to a named domain and use that](https://superuser.com/questions/152146/how-to-alias-a-hostname-on-mac-osx). 
 :::
 
-## Step 2) Frontend config
+## Step 2) Set Older Cookie Domain in the Backend Config
+
+:::caution
+- If `olderCookieDomain` isn't set, users with older sessions will get a 500 error from the session refresh endpoint, effectively locking them out. This will continue until `olderCookieDomain` is set correctly or they clear their cookies.
```
rishabhpoddar, line 118: this caution should come below the code snippet, cause when reading this, mentioning olderCookieDomain before explaining it doesn't make much sense.

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -104,15 +104,111 @@ init(
...
+)
+```
+
+</TabItem>
+</BackendSDKTabs>
+
+:::info
+- The `olderCookieDomain` value should be set to prevent clients from having multiple session cookies from different domains. This can happen when cookies from a previous domain are still valid and sent with requests. For instance, if your previous `cookieDomain` was `api.example.com` and the new one is `.example.com`, both sets of cookies would be sent to the apiDomain `api.example.com`, leading to an inconsistent state. This can cause issues until the older cookies are cleared. Setting `olderCookieDomain` in the configuration ensures that the SuperTokens SDK can automatically remove these older cookies.
+
+- Changing the `cookieDomain` can cause a temporary spike in requests, even if the `olderCookieDomain` is set correctly. This happens because older sessions, with older cookie domain, require additional refresh calls to clear their old cookies and set new ones. This spike is a one-time event and should not recur after the update.
```
rishabhpoddar, line 208: the spike point should be in a caution IMO

rishabhpoddar (changes requested):

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
anku255: Done

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
```diff
@@ -104,15 +104,111 @@ init(
...
 For local development, you should not set the `cookieDomain` to an IP address based domain, or `.localhost` - browsers will reject these cookies. Instead, you should [alias `localhost` to a named domain and use that](https://superuser.com/questions/152146/how-to-alias-a-hostname-on-mac-osx). 
 :::
 
-## Step 2) Frontend config
+## Step 2) Set Older Cookie Domain in the Backend Config
+
+:::caution
+- If `olderCookieDomain` isn't set, users with older sessions will get a 500 error from the session refresh endpoint, effectively locking them out. This will continue until `olderCookieDomain` is set correctly or they clear their cookies.
+
+- The value set for `olderCookieDomain` must be kept for 1 year because the cookie lifetime of the access token on the frontend is 1 year.
```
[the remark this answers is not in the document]
anku255: done

v2/session/common-customizations/sessions/multiple-api-endpoints.mdx
anku255: Moved it to caution.

rishabhpoddar (approved):

PR #798, opened 2024-05-29
fix: Update NestJS AuthGuard to use getSession

anku255:
## Summary of change
This PR updates the NestJS AuthGuard to use `getSession` instead of `verifySession`.

## Related issues
- Link to issue1 here
- Link to issue1 here

## Checklist

### DIFF
--- a/v2/emailpassword/nestjs/guide.mdx
+++ b/v2/emailpassword/nestjs/guide.mdx
@@ -290,9 +290,6 @@ export class SupertokensExceptionFilter implements ExceptionFilter {
     const ctx = host.switchToHttp();
 
     const resp = ctx.getResponse<Response>();
-    if (resp.headersSent) {
-      return;
-    }
 
     this.handler(
       exception,
--- a/v2/emailpassword/nestjs/guide.mdx
+++ b/v2/emailpassword/nestjs/guide.mdx
@@ -106,7 +106,7 @@ export class AuthModule implements NestModule {
 }
 ```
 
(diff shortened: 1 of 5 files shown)


v2/emailpassword/nestjs/guide.mdx
```diff
@@ -414,11 +400,16 @@ import { Session } from './auth/session/session.decorator';
...
 export class AppController {
   // ...
-  @Get('test')
+  @Get('/sessioninfo')
   @UseGuards(new AuthGuard())
-  async getTest(@Session() session: SessionContainer): Promise<string> {
-    // TODO: magic
-    return "magic";
+  getSessionInfo(
+    @Session() session: SessionContainer,
```
rishabhpoddar, line 406: the auth guard modifies the req object to add the session prop to it. How is that automatically picked up here?

Also, call this a test API instead of sessioninfo cause sometimes people think this is a necessary apis thats needed for supertokens integration

rishabhpoddar (changes requested):

v2/emailpassword/nestjs/guide.mdx
porcellus: Do we want to do the same thing in the cli example(s)? We could also add comments both here and there to explain that these are just example/tests endpoints.
The session prop is picked up by adding the Session decorator (defined elsewhere in this guide)

porcellus (approved):

v2/emailpassword/nestjs/guide.mdx
anku255: Changed it to `/test` and added a comment.

@porcellus  I didn't make this change in the CLI because the frontend code expects the `/sessioninfo` endpoint there.

PR #808, opened 2024-06-27
Chore: Send logs

ohansFavour:
## Summary of goal
Send sdk logs to the server when auth error occurs (either front_token_removed or auth_error).

## Summary of solution
Override the native console implementation

## Related issues
- Link to issue1 here
- Link to issue1 here

## Checklist

- [ ] Item1
- [ ] Item2

### DIFF
--- a/v2/src/components/httpNetworking.ts
+++ b/v2/src/components/httpNetworking.ts
@@ -1,5 +1,5 @@
 import axios from "axios";
-import { getAnalytics } from "./utils";
+import { getAnalytics, sendSDKLogsToBackend } from "./utils";
 
 export enum HTTP_REQUEST_ERROR {
     SESSION_EXPIRED,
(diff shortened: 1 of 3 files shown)


v2/src/components/utils.ts
```diff
@@ -106,4 +107,70 @@ export const recursiveMapAllChildren = (
...
+        }
+        logArray.push(data);
+
+        localStorage.setItem(SDK_LOGS_STORAGE_KEY, JSON.stringify(logArray));
+    } else {
+        oldConsoleImplementation(data);
+    }
+};
+
+export async function sendSDKLogsToBackend() {
```
rishabhpoddar, line 148: no need for this function since it's not being used.

rishabhpoddar (changes requested):

v2/src/components/httpNetworking.ts
```diff
@@ -205,6 +205,7 @@ function sendAnalyticsIfFrontTokenRemoved(url: string, frontTokenExists: boolean
             url,
             headers
         });
+        await sendSDKLogsToBackend()
```
ohansFavour, line 208: @rishabhpoddar we use it here

